PR #5, opened 2026-08-31
fix(aerospace): quote app-name placeholders so two-word names work

jborlum:
`read_apps` accepts spaces in app names (`[A-Za-z0-9 ._-]`), but `exec-and-forget` runs through bash — so an unquoted two-word name word-splits and the launch silently fails:

```
BROWSER=Google Chrome  ->  open -a Google Chrome  ->  Unable to find application named 'Google'
```

Same for `Visual Studio Code`, `IINA Advance`, etc. `omacosy-karabiner-omniwm` already quotes these, so this just makes the AeroSpace path match.

Tested on macOS 26.6.2 / AeroSpace 0.21.3 with `BROWSER=Google Chrome`; single-word names unaffected.


paulsp94 (approved):
LGTM good catch

PR #8, opened 2026-09-01
feat(bar): support custom workspace icons

### LINKED ISSUE #4: Allow optional custom icons for workspaces
Workspaces currently show an app icon when they contain exactly one app, and fall back to their workspace number otherwise.

   Could omacosy support optional per-workspace icons? This would keep semantic workspaces recognizable when they are empty or contain multiple apps.

   Suggested behavior:

   1. Custom workspace icon, when configured
   2. Sole-app icon
   3. Workspace number

   The mappings could be loaded once when the bar starts, so drawing remains process-free and event-driven.

   For multi-monitor setups, a slot mapping such as `4` could apply to both workspaces `4` and `14`.

   I’m willing to submit a focused PR. Would you prefer the configuration to use image paths, app bundle IDs, or Nerd Font glyphs?

Jonathanm10:
## Why

Workspaces currently lose their semantic icon when they are empty or contain more than one app. This adds the optional startup-loaded mapping discussed in #4 without putting file access or application lookup on the draw path.

## Scope

- Read `~/.config/omacosy/workspace-icons.conf` as `workspace=value` entries.
- Support one-scalar Nerd Font glyphs and installed application bundle identifiers.
- Prefer exact workspace names, then single-digit shorthand for multi-display numeric workspaces.
- Keep the existing sole-app icon and workspace-number fallbacks.
- Preserve user-authored files under `~/.config/omacosy` during uninstall.
- Document configuration, precedence, diagnostics, and restart behavior.

## Tradeoffs

The bar reads this configuration only at startup. Applying an edit requires restarting the launch agent, which avoids a file watcher and keeps rendering event-driven.

## Blast Radius

The runtime change is limited to workspace-chip content selection in `helper/bar.swift`. `install.sh` does not create or overwrite the optional file. `uninstall.sh` now removes only installer-owned files before attempting to remove the config directory.

## Verification

- Production-equivalent optimized Swift build with the linked private frameworks and embedded Info.plist.
- `/bin/bash -n uninstall.sh`.
- `git diff --check origin/main...HEAD`.
- Live stacked-bar run on macOS verified configured glyphs, bundle icons, duplicate handling, malformed-line diagnostics, and unavailable-bundle fallback before the follow-up parser edge-case commit.
- The final head was rebuilt after the parser and uninstall edge-case commit.

Closes #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,38 @@ clickable (app menus were unreachable before that fix). It drops back
 behind everything when the pointer leaves. Under OmniWM the bar simply
 stays visible in its reserved strip and never plays this game.
 
+### Workspace icons
+
+You can set workspace icons in the optional
+`~/.config/omacosy/workspace-icons.conf` file. Each non-comment line has one
+workspace name, an equals sign, and either one Unicode scalar or a reverse-DNS
+application bundle identifier.
+
+```
+1 = ★
+14 = ◆
+
+4 = com.apple.Safari
+```
+
+The bar first uses a configured icon. It then uses the icon of the sole app on
+that workspace, and finally shows the workspace's last digit. Exact workspace
+names win. In this example, workspace `14` uses `◆`, not the `4` shorthand.
+The shorthand applies only to multi-digit, all-numeric workspace names ending
+in `1` through `9` when they have no valid exact declaration.
+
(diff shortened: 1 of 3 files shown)


paulsp94 (approved):
LGTM, nice work

PR #11, opened 2026-09-03
fix(borders): the focus ring keeps up with same-app focus switches

hippohide:
## Problem

Moving focus between two windows of the same app (e.g. `focus left` in AeroSpace) leaves the ring on the old window for a noticeable beat — measured at a median of **225ms**, up to **507ms**, before it catches up.

Two causes, both specific to a same-app switch, where the focused window is *raised* but nothing moves, nothing is created, and the front application does not change:

**1. The reorder is never announced.** The daemon subscribes to `FRONT_CHANGE`, but that event lands ~16ms *before* the window server applies the raise, so the tick it triggers still resolves the old topmost window. The events that announce the reorder itself — `WINDOW_ORDER` (808) and `WINDOW_VISIBILITY` (815) — are not subscribed, so nothing wakes the daemon afterwards and the update falls through to the 0.5s heartbeat. `bar.swift` already subscribes to both.

**2. The mid-storm shortcut re-confirms the stale window.** `focusedWindowFrame()` skips the full window-list walk mid-storm and re-queries `lastWid` directly for up to 250ms. It only checks that `lastWid` is still a live, layer-0 window owned by the frontmost app — all of which stays true of the window focus just moved away from within one app. So the old window keeps passing the check and stays ringed until the rescan window expires.

This was the larger half: subscribing to the new events alone changed nothing.

## Solution

- Subscribe to `WINDOW_ORDER` and `WINDOW_VISIBILITY`, so the raise itself wakes the daemon.
- Add a `focusMayHaveChanged` flag, set by every event except move/resize, that forces a full rescan instead of trusting `lastWid`. Drags keep the shortcut they were built for.
- Cache `omniwmActive()` for 0.5s — it walks every running app (0.36ms) twice per tick, on both WM paths. This one is an independent drag-smoothness win rather than part of the latency fix; happy to split it out or drop it if you'd prefer the PR scoped to the bug alone.

## Measured

Time from the window server applying the reorder to the ring window moving, 8 samples:

|        | before | after |
|--------|--------|-------|
| median | 225ms  | 11ms  |
| max    | 399ms  | 27ms  |

Total keypress-to-ring is now ~95ms, the remaining ~70ms being AeroSpace's own CLI/AX round-trip.

Tested on macOS 26.3, AeroSpace, two displays (one notched built-in, one external). The event-ordering claims are timing-sensitive and OS-version-specific, so they're worth a second look on other setups.

### DIFF
--- a/helper/borders.swift
+++ b/helper/borders.swift
@@ -6,8 +6,9 @@
 //
 // Event-driven via private SkyLight window-server notifications (the
 // same layer JankyBorders and yabai use, verified on macOS 26.3):
-// front-app changes, window create/destroy, and per-window move/resize
-// all arrive as callbacks the instant the WindowServer processes them.
+// front-app changes, window raise/reorder, window create/destroy and
+// per-window move/resize all arrive as callbacks the instant the
+// WindowServer processes them.
 // A 0.5s heartbeat remains as a safety net for anything eventless,
 // and kqueue watches cover theme/config changes. Notify procs only
 // fire while the connection's event port is drained — see the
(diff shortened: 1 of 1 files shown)


paulsp94 (approved):
LGTM

PR #12, opened 2026-09-04
perf(spawn): Ghostty windows come from the running instance

jborlum:
`open -na` starts a second Ghostty app instance for every Super+Enter. Ghostty 1.3 ships a scripting dictionary, so the window can come from the instance already running instead — the same window for a fraction of the work.

- window open: ~300 ms -> ~140 ms
- `omacosy-spawn Ghostty` end to end: ~0.55 s -> ~0.35 s
- three presses back to back: 2.34 s -> 1.65 s

open(1) stays the path for every other app, for a Ghostty that is not running (addressing an app that is down would launch it *and* let it open its own initial window, so two windows for one press), and for Ghostty before 1.3, which has no dictionary and raises. The new window's directory is pinned to `$HOME`, which is where `open -na` starts one, so no behaviour rides along with the speed-up.

The lock and the split-hint wait are untouched — they just hold for less time now.

Tested on macOS 26.6.2 (25G83), single display, AeroSpace path: `bash -n` clean, run under stock `/bin/bash` 3.2.57, windows tile and focus as before, and the fallback confirmed by spawning a name that does not resolve. I do not run OmniWM, so that branch is unexercised beyond the shared `spawn_app` call.


paulsp94 (approved):
LGTM great work, but seems to have a conflict with you other PR I already merged 
#13 
GH has a stacked PR feature that makes these PRs easier to merge one after another might be worth a look

jborlum:
> LGTM great work, but seems to have a conflict with you other PR I already merged #13 GH has a stacked PR feature that makes these PRs easier to merge one after another might be worth a look

Thank you. Good point. I've just been customizing and benchmarking a bit on my side. Didn't think too much about the stacking. Will fix.

paulsp94:
Hey @jborlum do you wanna do the rebase or do you want me to pick it up from here?

jborlum:
I have not had time! Feel free to pick it up. Thank you.

paulsp94:
Picked this up and got it merge-ready. Your commit is unchanged apart from the rebase; the second commit is mine.

**Rebased onto main.** The only conflict was `config/ghostty/config`: #13 landed `resize-overlay = never` directly above the comment block this PR rewrites. Both are kept.

**The one thing that worried me, now settled.** The scripting dictionary has a separate `activate window` command, so it was not obvious that `new window` fronts the app — and the AeroSpace path's wait depends on it, since the hint only stamps the state file when `on-focus-changed` fires. Measured from a non-Ghostty front app: after `new window`, `lsappinfo front` reports Ghostty. So the wait is fine, and nothing extra is needed.

**Added a timeout.** The Apple event runs while the spawn lock is held, and an event to a wedged app waits two minutes by default — that would take Super+Enter down with it. `with timeout of 5 seconds`, and the error lands in the same `open(1)` fallback as a missing dictionary.

**Re-measured, timed to a window the WM can see rather than to the command returning.** Your numbers are kept alongside mine, since they are different machines.

| path | first window | ghostty processes after 3 spawns |
| --- | --- | --- |
| dictionary | ~225 ms | 1 |
| `open -na` | ~810 ms | 4 |

That process column is the other half of the win, and it is what the config note claims.

**End-to-end on the OmniWM branch you could not exercise**, live: Ghostty down, first spawn takes `open(1)` and lands one window; the next two come from the running instance at 264 ms and 253 ms through the script including its own wait; window count grows by exactly one each time; process count steady at one throughout.

**One note rather than a change:** the app name is matched literally, so a renamed Ghostty bundle quietly takes the `open(1)` path. Slower, never wrong. I wrote that into the comment instead of adding a bundle-id lookup, which would cost more on the hot path than the special case saves.

Not verified here: a live AeroSpace session, since this machine runs OmniWM. The activation question above was the part of that path that actually mattered.

PR #19, opened 2026-09-11
feat(tiling): a lone window can take the whole display, like Super+F

### LINKED ISSUE #16: Feature request: fullscreen a workspace holding one window
Hardware: Macbook Air M1 (2020 - no notch)
OS: MacOS 27 (Golden Gate Beta)
Config (Workspace/Tiling): Aerospace

---

Video:

https://github.com/user-attachments/assets/6ed4ec8e-308c-4619-9136-ea655e0b7b3e

---

## The idea

When a workspace holds exactly one window, give that window the whole
display with no gaps. When a second window appears, drop back to normal
tiling. When you close back down to one, go full again.

Nothing to press, and the SUPER + F shortcut still works. It follows what is in the workspace.

Hyprland does not do this by default either, so this is a request rather
than a missing parity feature. It comes up because a single window with
gaps on all four sides wastes space that has no other claimant, and
`Super+F` by hand gets tedious when you open and close windows often.
The goal is to have a fully immersive and distraction-free experience.
Neon borders are nice but sometimes you just want to see what's on screen and nothing else, opionated as always. When there is only one window left, what's the point of highlighting it, it's the main focus.

## How much it is worth

Honestly, it depends on your gaps. With `outer` gaps of 8 it buys 8px per
side, which is cosmetic. With the shipped 42px top gap on a flat monitor
it is more noticeable. Someone running larger gaps deliberately would
want it off.

So: opt-in, off by default.

## What I am running

A shell helper on AeroSpace's `on-focus-changed` hook, alongside
omacosy's own `split-hint` call:

```toml
on-focus-changed = [
  'exec-and-forget $HOME/.local/bin/omacosy-helper split-hint',
  'exec-and-forget $HOME/.local/bin/aerospace-auto-fullscreen',
]
```

```sh
#!/bin/sh
apply() {
  n=$(aerospace list-windows --workspace focused 2>/dev/null | wc -l | tr -d ' ')
  [ -z "$n" ] && return 0
  if [ "$n" = "1" ]; then
    aerospace fullscreen on --no-outer-gaps 2>/dev/null
  else
    aerospace fullscreen off 2>/dev/null
  fi
}
apply
```

## Three things that were not obvious

Anyone implementing this will hit all three, so they are worth stating.

**1. `aerospace fullscreen` with no argument is a toggle.** That is how
`Super+F` works. Called from a hook without an explicit `on` or `off`, it
flips the state on every focus change instead of setting it, so whether a
workspace ends up fullscreen depends on how many focus changes have
happened. It looks random. Always pass `on` or `off`.

**2. A newly opened or closed window is not always in the window list
yet** when the focus hook fires, so the first pass counts wrong. The
first window on an empty workspace is exactly that case, and it stayed
tiled until a second window forced another focus change. A delayed second
pass fixes it.

A lock file is the obvious guard against passes stacking up, and it is
wrong: closing a window fires several focus changes at once, the second
run finds the first run's lock and skips its own retry, so the retry is
suppressed exactly when it is needed. A generation token works — every
run schedules a retry, only the newest one acts.

**3. It fights `Super+F`.** On a solo workspace the helper sets
fullscreen on; you press `Super+F` to leave it; the next focus change
sets it back within a second. The reverse too: `Super+F` in a multi-window
workspace is cancelled because the helper sets `off` whenever it counts
more than one.

I handle that by recording the workspace and window count when `Super+F`
is pressed, and having the helper skip while both still match. The record
has to be **one-shot**, deleted the first time the state differs, or it
comes back to life: press `Super+F` on a solo workspace (record "1 1"),
open a window (count 2, record ignored), close it (count 1 again, record
matches *again*) and the workspace never returns to fullscreen.

## Costs

- A subprocess on every focus change, on top of the existing `split-hint`
  one, plus a delayed second pass.
- Under OmniWM none of this applies; it has its own layout handling.

## What a native version could do better

Omacosy already knows the window count per workspace for the bar and the
overview, so a built-in version would not need `aerospace list-windows`
at all, and could react to a window-count change rather than to every
focus change. That removes the subprocess, the delayed pass and the
`Super+F` conflict in one go, since a deliberate `Super+F` and an
automatic change would no longer be indistinguishable.

## Considerations

This is obviously a better feat for non-notched Macbooks and desktops owners, so for those who have the notch and/or want to always show the menu bar then some adjustment should be applied. I do not own a notched Macbook, so I cannot test.

paulsp94:
Closes #16.

## What

Opt-in with `omacosy-solo-fullscreen on` (off by default, `off` or no-arg toggle to undo). Under AeroSpace, a workspace holding **one tiled window** shows it the way `Super+F` does: the whole display, no gaps, no ring, the bar tucked away until the pointer touches the top edge (notch strip blacked out on notched displays, same as `Super+F`). A second window drops back to normal tiling; closing down to one goes full again.

Under OmniWM there is nothing to switch: dwindle's `singleWindowFit = "fill"` already gives a lone window the frame its own `Super+F` uses, so the command records the setting and says so.

## How

`bin/omacosy-solo-fullscreen sync` runs as a second `on-focus-changed` entry. Opted out, it exits before asking AeroSpace anything.

It acts on **transitions, not levels**, which is what keeps it from fighting `Super+F` (the three pitfalls from the issue):

- **No toggle fights.** Each visible workspace remembers the window it last saw alone; only a change acts. Explicit `fullscreen on|off`, never the toggle.
- **Hand-made fullscreens are never touched.** A window counts as ours only when `fullscreen on --fail-if-noop` actually changed something, and `off` hands back only those.
- **Window-list lag.** One re-check 0.5 s later catches a closing window still listed or a first window not yet listed. Only the newest run in a burst re-checks (generation token, not a lock, so a burst cannot swallow the retry that mattered).
- Floats do not count. Both displays' visible workspaces are judged in one `list-windows --workspace visible` call.

Plain shell like the other layout tools, so no helper rebuild or re-signing.

## Files

- `bin/omacosy-solo-fullscreen`: the script
- `config/aerospace/aerospace.template.toml`: second `on-focus-changed` entry
- `install.sh` / `uninstall.sh`: link and remove the script and its marker
- `README.md`: "One window, whole display" under Tiling

## Testing

- Stub AeroSpace harness, 29 scenarios, passing under bash 5.2 and macOS `/bin/bash` 3.2: opt-out makes no calls; `Super+F` off on a lone window stands; manual fullscreen on a busy workspace survives closes and new neighbours; unfocused arrival hands ours back; floats ignored; two displays judged independently; lagging list on close and on first window; burst retries once; `off` hands back only ours; toggle.
- Template parses as TOML; the bar's cheatsheet parser only reads `[mode.main.binding]`, so the multi-line array is ignored.
- `on` / toggle-off / bad-argument paths run cleanly live under OmniWM.

**Not yet verified against a live AeroSpace session**, and the 0.5 s retry delay is a judgment call rather than a measurement. Existing AeroSpace installs pick up the hook on the next `install.sh` / `omacosy-update` run.

https://claude.ai/code/session_01VaidEMMJiYMGJ8QYBVm1YZ


datarip:
You asked me to check it out and say if it feels good, so here it is properly.

First the awkward part. On #16 I lacked the confidence to do this myself: I am
not a developer by trade and I had never worked in a Swift codebase. Then I found
some courage, motivated myself, and built it anyway — without telling you. That
is my fault and I am sorry, because it meant you worked on it too. The duplicated
work is on me, not on you.

**I am not asking you to drop #19.** I have taken three things out of it into
mine, because on each one you are right and I was wrong:

- **`fullscreen on --fail-if-noop`.** A window counts as the rule's own only when
  the call actually changed it. I was inferring the same thing from remembered
  window ids, which is more state and more ways to be wrong: the snapshot can say
  a window is not fullscreen, the user can press `Super+F` in the gap before the
  call, and the rule then claims a window it never touched — and hands it back on
  `off`.
- **`off` releases only what the rule took**, so a fullscreen made by hand
  survives switching the feature off.
- **A real `on | off | toggle` command** instead of a config file. I have dropped
  my `config/fullscreen.conf` entirely; the marker file is the whole setting, as
  in yours.

So the interface argument does not need having. Yours is better and I have
adopted it. What I want to make a case for is **the trigger**.

## One defect in #19, worth fixing whatever else you decide

```awk
NF >= 3 && $3 != "floating" { print $1 "|" $2 }
```

Only `floating` is excluded, but `%{window-layout}` has six values:

```
tiling (reported as h_tiles / v_tiles / …)
floating
macos_native_fullscreen
macos_native_minimized
macos_native_window_of_hidden_app
NULL-WINDOW-LAYOUT
```

A `Cmd+H`'d app keeps its window in AeroSpace's list as
`macos_native_window_of_hidden_app`, so a workspace **showing** one window counts
two and never goes full. Two windows are in that exact state on this machine as
I write this, with no test set up:

```
$ aerospace list-windows --all --format '%{workspace}|%{window-layout}|%{window-id}|%{app-name}'
4|macos_native_window_of_hidden_app|1891|Ghostty
4|macos_native_window_of_hidden_app|1894|Ghostty

$ osascript -e 'tell application "System Events" to get visible of process "Ghostty"'
false
```

It is in AeroSpace's source too, in `normalizeLayoutReason.swift`: a window is
rebound to `workspace.macOsNativeHiddenAppsWindowsContainer` when the app is
hidden, `automatically-unhide-macos-hidden-apps` is off (which is the default),
and the window is not fullscreen or minimized. The arm explicitly handles a
window that was previously tiled.

One extra word in the filter fixes it. I hit it with Notes hidden behind an
editor.

**Minimized windows are fine in both of ours**, in case it looks like the same
problem — AeroSpace drops them from `list-windows` entirely, so they cannot be
miscounted. I checked, because I assumed the opposite first.

## What the focus trigger costs

I ran your script with a stubbed `aerospace`, so these are its own numbers.
MacBook Air M1, 1440x900, AeroSpace 0.21.3-Beta, macOS 27.0.

- Opted **out**: 8.8 ms per focus change. Small, but every user pays it forever,
  including everyone who never turns the feature on.
- Opted **in**: 0.59 s per run, of which 0.5 s is the deliberate re-check sleep.
  Two real `list-windows` calls add ~42 ms, and each run forks about ten
  awk/grep/subshell processes.

The part that may not show on your machine: with focus-follows-mouse every hover
is a focus change. I counted 12 in 10 seconds of ordinary pointer movement.

## Five things a focus hook cannot reach

| | #19 |
| --- | --- |
| after sleep and wake | does nothing, and cannot |
| after boot | waits for the first focus change |
| `Super+T` floats one of two | `on-focus-changed` fired 0 times |
| window opens on a background workspace | `--workspace visible`, so it stays tiled until you go there |
| **a minimized window is restored** | **AeroSpace emits no event at all** |

The fourth is a one-word change: `--all` instead of `--workspace visible`.
`fullscreen on --window-id` works on a window that is not on screen, so it costs
nothing. I measured 260 ms from the window appearing to it being full, in the
background, against "tiled until the workspace is next visited" — which is most
windows, since app-to-workspace rules put almost everything somewhere else.

**The fifth one is new and I found it while testing this.** Restore a minimized
window and AeroSpace re-adopts it onto the *focused* workspace. Running
`aerospace subscribe --all --no-send-initial` across the whole cycle emits
**zero lines** — no `window-detected`, and no `focus-changed` either, so
`on-focus-changed` has nothing to fire on. A workspace silently goes from one
window to two and #19 cannot know. I have reported the underlying behaviour to
AeroSpace: nikitabobko/AeroSpace#2263.

My daemon does see it, because it watches SkyLight rather than AeroSpace's
events. From its own log, during that test:

```
02:04:58  released 1 window(s) that are no longer tiled     <- minimized
02:05:03  3 holds 2, window 1643 off                        <- restored onto ws 3
```

I am not claiming that as a design win I planned. I picked SkyLight because
`aerospace subscribe` announces a window *arriving* and not a window *leaving*,
and closing down to one window is the event this rule exists for.

The first row is the structural one, and it is the honest reason a second
implementation exists at all. Waking fires no focus change and no window event.
AeroSpace re-detects every window — I saw five inside 10 ms — places them all on
the focused workspace, and fullscreen is dropped everywhere. Only `NSWorkspace`'s
sleep and wake notifications see any of it, and **only a live process receives
those**. A manual `Super+F` is implied by no window count, so it has to be
RECORDED before the sleep rather than recomputed after it. That is my omission
too: I found both sleep problems after filing #16 and never posted them.

## What I am proposing

`helper/solo.swift`, a small daemon under launchd, with **your command as its
front end**. `omacosy-solo-fullscreen on|off|toggle|status` is the whole user
interface; the daemon is what the marker switches on.

- KeepAlive is `SuccessfulExit=false`, so a crash restarts and a deliberate exit
  stays exited. **Switched off, nothing runs at all** — not a hook, not a
  process.
- Its own binary rather than part of `omacosy-bar`: a Swift trap cannot be
  caught, and nothing about fullscreening a window should be able to take the
  menu bar down.
- 35 MB resident, against 33 for `omacosy-ffm` and 39 for `omacosy-borders`. 0%
  CPU idle. One `list-windows` per window event, 21 ms. `on` to the window being
  fullscreen: 0.57 s.
- Thirteen assertions driven from a script against real AeroSpace, twice in a row
  in one process, including the two that matter most here: the rule does not
  claim a fullscreen made by hand, and `off` does not take it away. Sleep
  verified across a lid close that really did damage — 1 of 3 fullscreens
  survived the wake, and all 3 were restored.

## The collision, said out loud

Mine ships `bin/omacosy-solo-fullscreen` — **the same path and the same command
name as yours**, because I deliberately adopted your interface. Git will see two
different files at one path and there is no clean automatic merge. That is not a
thing to discover in a diff, so: whichever way this goes, one of the two files
has to win outright, and I am happy for the answer to be yours.

## Two things I cannot claim

I **cannot test a notched display** — this machine has no notch — so I would not
want to claim that part works. I raised this on #16 for your implementation and
it applies to mine identically.

You wrote that #19 is not yet verified against a live AeroSpace session and that
the 0.5 s retry is a judgment call. **I have a live AeroSpace machine and I am
happy to run #19 on it and report back**, the same way I did for #20, if that is
useful to you. Say the word.

## Where I actually stand

I am happy with any answer. If you would rather keep the shell script and have
only the part a hook cannot reach, say so and I will rebuild mine as a trigger
that calls your script. If you want none of it, **#19 plus the hidden-app word is
a good feature** and I will keep mine in my fork and stop taking up your time.

paulsp94:
Pushed a fix for the defect you found, plus two more it led to.

**The layout filter.** Confirmed exactly as you described: the installed 0.21.3-Beta binary carries `macos_native_fullscreen`, `macos_native_minimized`, `macos_native_window_of_hidden_app` and `NULL-WINDOW-LAYOUT` beyond the five in the man page. Excluding only `floating` broke it both ways, as you said. Tiles are allowlisted now: `*_tiles` and `*_accordion`.

**Ownership was recorded before the call that earns it.** One refused call, AeroSpace busy mid-relayout, left that window tiled until the count crossed again. The record is written once the window really is fullscreen. The listing now carries `%{window-is-fullscreen}`, so a window already fullscreen by hand is recorded but never claimed, without a second call.

**An emptied workspace kept its record**, which blocked that window's return. Records are pruned against the live listing, and the ours-list is pruned to windows that still exist — AeroSpace reissues window ids, and a dead id could have handed back a stranger's fullscreen on `off`.

Two smaller ones: the listing covers every workspace rather than the visible ones, so a hidden workspace is already right when you arrive; and the marker test moved into the hook with `exec`, so opting out costs nothing beyond the shell `exec-and-forget` spawns anyway.

29 stub-driven checks pass under bash 5.2 and `/bin/bash` 3.2, including hidden-app, minimised, accordion, persistent-refusal, emptied-workspace and id-reuse cases.

Your #31 is still the better answer on two points I have not taken here: it excludes hidden-app windows by construction and it survives a wake without a hook firing. I will treat those separately rather than quietly absorb them.

PR #20, opened 2026-09-11
fix(tiling): a refocus read is checked, so a close cannot poison the dwindle chain

### LINKED ISSUE #17: Bug: split-hint: wrong dwindle direction after a window closes (with fix)
Hardware: Macbook Air M1 (2020 - no notch)
OS: MacOS 27 (Golden Gate Beta)
Config (Workspace/Tiling): Aerospace

## Issue

I noticed that when opening a second window in the same workspace, such as hitting `SUPER + ENTER` (terminal) or `SUPER + SHIFT + F` (new Finder window), the new window tile randomly opens horizontally instead of vertically like expected.

## Summary

`split-hint`'s `read` branch trusts the focused window's frame whenever a
state file exists. After a window **closes**, that assumption breaks: the
survivor takes focus with a lower window id, misses the `predicted`
branch, and lands in `read` before AeroSpace has re-expanded it. The
frame read is the survivor's old half-size slot, so the direction comes
out wrong and the next window opens below instead of beside.

Roughly half of `SUPER + SHIFT + F` presses after a close are affected, and in a lower occurrence `SUPER + ENTER` too.

A two-line fix is below. I have been running it and the wrong splits are
gone.

## Environment

- macOS 26, Apple Silicon M1, no notch
- Single built-in display, 1440x900
- AeroSpace (default), not OmniWM
- omacosy at `59b5202`
- Default gaps except `outer.top`, no other local changes to
  `helper/main.swift`

## Steps to reproduce

1. Empty workspace.
2. `Super+Enter`. The terminal fills the workspace.
3. `Super+Enter`. The second window opens to the right. Correct.
4. `Super+W` to close the second window. The first re-expands.
5. `Super+Enter`.

Expected: the new window opens to the **right**, because the survivor is
1424x883, wider than tall.

Actual: about half the time it opens **below**.

## Cause

In `helper/main.swift`, the `read` branch:

```swift
} else if state != nil, let f = frame() {
    // an existing window refocused mid-burst: its frame is settled
    (w, h) = (f.2, f.3)
    how = "read"
}
```

The comment holds for the case it was written for, a hover or keyboard
refocus mid-burst. It does not hold after a close:

- The closed window's hint stays in the state file for the 3s TTL.
- The survivor takes focus. Its id is **lower** than the state's, so
  `wid > s.wid` is false and the `predicted` branch is skipped.
- `state != nil` is true, so `read` runs.
- AeroSpace has not re-expanded the survivor yet, so `frame()` returns
  its pre-close slot.

## Log evidence

`wid=2120` is the first terminal,
whose true size alone on the workspace is 1424x883.

Correct, read at its real width:

```
1788865910.136849 wid=2120 1424x883 (static)  -> h rc=0
1788865917.910901 wid=2226  708x883 (settled) -> v rc=0
```

Wrong, read at the half-width slot it had before the other window closed:

```
1788865891.995008 wid=2202 708x883 (settled) -> v rc=0
1788865892.996736 wid=2120 708x883 (read)    -> v rc=0
```

The same window appears at three sizes across one session, which are
three layout states of one window:

```
wid=2120 1424x883   alone on the workspace
wid=2120 1424x495   stacked, half height
wid=2120  708x883   side by side, half width
```

Which state `read` catches decides the split.

## A second, larger case: the window server lags

`chainAlive` below fixes the case where the state's window is gone. It is
not sufficient, and the reason matters for anyone attempting this fix.

**`CGWindowListCopyWindowInfo` still returns an entry for a window that
has just closed.** So a liveness test based on the window server returns
true for a dead window, and the `read` branch runs anyway.

Reproduced with open, close, open, close on a workspace whose only window
is 1424x883:

```
wid=2390  708x883 (read) -> v
wid=2390 1424x495 (read) -> h
wid=2390  708x883 (read) -> v
```

Neither 708x883 nor 1424x495 is that window's real frame. Both are
leftovers from when the Finder window was still there. Sampling the frame
twice does not catch it either, because AeroSpace has not *started*
re-expanding yet, so two samples 50ms apart agree on the same stale
value.

**AeroSpace knows what it placed; the window server does not yet.** If
the focused workspace holds exactly one window, that window's frame must
fill the display minus gaps. A frame much smaller than that is provably
stale, whatever the window server says.

Note that width alone is not enough: 1424x495 is full width on a
1440-wide display and still stale. Both dimensions have to be checked.

## Why `Super+Enter` does not show it

`omacosy-spawn` blocks until the hint has stamped the state file, which
spaces presses far enough apart that the 3s TTL expires and the settle
path runs. `open ~` returns immediately, so presses land inside the TTL,
the chain stays alive, and `read` is taken every time. The bug is in
`read`, not in the bindings.

## Fix

Skip the `read` branch when the window the state refers to no longer
exists. A dead chain means a layout change is in flight, so the frame has
to be settled rather than read. The existing settle path already handles
this correctly, because a re-expanding survivor **moves**, which is what
its `moved` check waits for.

Add next to `frame()`:

```swift
    func chainAlive(_ id: UInt32) -> Bool {
        guard let l = CGWindowListCopyWindowInfo(.optionIncludingWindow, id)
            as? [[String: Any]] else { return false }
        return !l.isEmpty
    }
```

Then change the branch condition:

```diff
-    } else if state != nil, let f = frame() {
+    } else if let s = state, chainAlive(s.wid), let f = frame() {
```

And, because the window server lags, refuse a read the layout
contradicts:

```swift
    func soloAndFull(_ f: (CGFloat, CGFloat, CGFloat, CGFloat)) -> Bool {
        let bin = ["/opt/homebrew/bin/aerospace", "/usr/local/bin/aerospace"]
            .first { FileManager.default.isExecutableFile(atPath: $0) } ?? "aerospace"
        let pr = Process()
        pr.executableURL = URL(fileURLWithPath: bin)
        pr.arguments = ["list-windows", "--workspace", "focused"]
        let pipe = Pipe()
        pr.standardOutput = pipe
        pr.standardError = FileHandle.nullDevice
        guard (try? pr.run()) != nil else { return true }
        let out = pipe.fileHandleForReading.readDataToEndOfFile()
        pr.waitUntilExit()
        let count = String(data: out, encoding: .utf8)?
            .split(separator: "\n").filter { !$0.isEmpty }.count ?? 0
        guard count == 1 else { return true }
        let display = CGDisplayBounds(CGMainDisplayID())
        return f.2 >= display.width - 120 && f.3 >= display.height - 120
    }
```

```diff
-    } else if let s = state, chainAlive(s.wid), let f = frame() {
+    } else if let s = state, chainAlive(s.wid), let f = frame(),
+              soloAndFull(f) {
```

After both, the survivor logs as `(settled)` at its full size and the
direction is correct.

A subprocess here goes against the design: the
comments make clear that avoiding one is why `predicted` exists. It is
only reached when the chain is alive, so bursts of opens never pay it,
but there is very likely a cheaper way to ask AeroSpace what it has
placed that I do not know about.

### Implementation Cost

One extra `CGWindowListCopyWindowInfo` call, plus one `aerospace
list-windows` subprocess, about 30ms, on the `read` path only.

Latency only in the case that was broken: a close now falls into the
settle loop, up to about 1.2s worst case and usually far less. Personally I noticed no slowdowns on my end. Bursts of
opens are untouched, because they use `predicted`. Hover and keyboard
refocus with a live chain still use `read` and are as fast as before.

### An alternative, if the extra call is unwanted

When the workspace holds exactly one window, that window's frame must
equal the workspace's usable area, so the direction could be computed
from the workspace instead of the frame. That avoids both the extra call
and the settle wait, but it needs the usable area, which `split-hint`
does not currently have.

### Testing done

Single display, AeroSpace, one machine. Reproduced the failure reliably
before the change and not at all after it, over a few dozen open/close
cycles. Not tested on multiple displays, and not tested under OmniWM.

## Attachments

[main.swift.diff.txt](https://github.com/user-attachments/files/31966597/main.swift.diff.txt)
- `main.swift.diff`, the change above as a patch against `59b5202`.

## Note on scope

Stock config: `enable-normalization-flatten-containers` is `false`, the shipped value. I have not changed it.

Mentioning it because the log shows the direction being chosen wrongly by the hint itself, before AeroSpace places anything. Whatever AeroSpace does with the tree afterwards is downstream of a `split vertical` that should have been `split horizontal`.

paulsp94:
Closes #17. Thanks @datarip for the diagnosis, the logs and the patch that pinned this down.

## The bug

`split-hint`'s `read` branch trusts the focused window's frame whenever a fresh state file exists. After a close, the survivor refocuses with a **lower** window id, skips `predicted`, and lands in `read` before AeroSpace has re-expanded it. The frame is its old half slot, so the hint splits the wrong way, and it stamps that stale size into the chain, where the next window's prediction halves it again.

## Why not the patch in the issue

- `chainAlive` cannot tell: the window server still lists a just-closed window (the issue says as much).
- `soloAndFull` only covers closing down to **one** window. Close one of three and the survivor re-expands from a quarter to a half slot and still reads stale. A window moved or hidden away has the same shape, and it is alive anyway.
- It measures against the main display, so on a second display the check always drops to the slow path.
- It puts an `aerospace list-windows` subprocess on the read path, which the `predicted` design exists to avoid.

## The fix

Keep the read fast, then check it. After issuing the read-based split and stamping the state, the helper watches the window's frame for the **existing** settle grace (5 × 75 ms). If the frame changed:

- the state is corrected to the settled size, because the next window's prediction halves this slot, so a changed size matters even when the direction holds;
- the split is issued again if the direction changed.

Issuing it again is invisible and safe. After the first `split` the window is its container's only child, and per `man aerospace-split`, `split` on an only child just changes that container's orientation. That holds only while no window has joined, so **any newer hint in the state file cancels the correction**, since a second split would nest the newcomer.

Cost: a hover or keyboard refocus with nothing moving gets the same hint at the same moment; the process just lives ~375 ms longer doing a handful of in-process `CGWindowList` reads. No new subprocess unless a correction is due. `predicted` and the settle fallback are unchanged, moved into small local functions (`settle`, `direction`, `split`, `note`).

The hint log gains `(re-read)` lines; `rc=-` means the direction held and no second split was issued.

## Testing

Driven against a real `CGWindowID`: a click-through panel whose size changes on a schedule, with AeroSpace's server down, so `split` fails fast (`rc=2`) and the decision, the correction and the state file are what is under test. 19/19 checks pass.

- **#17 case, main:** survivor at 708x883 re-expands to 1424x883 after the read → `(read) -> v`, never corrected, chain left at 708x883. **Reproduces the bug.**
- **#17 case, this branch:** `(read) -> v` then `1424x883 (re-read) -> h` with a second split; chain corrected to 1424x883.
- **Size changes, direction holds** (1424x495 → 1424x883): `(re-read) -> h rc=-`, no second split, chain corrected.
- **Newer hint lands, then the window shrinks** (a sibling joined): no correction, newer hint left in place.
- **Refocus, nothing moves:** a single `(read)` line, no correction.
- **New window in a burst (`predicted`):** unchanged, exits in ~90 ms.
- **No chain, static window (settle fallback):** unchanged.

**Not verified against a live AeroSpace session** (the test machine runs OmniWM). The fix assumes AeroSpace starts re-expanding the survivor within the settle grace; the issue's own testing showed the settle path catching exactly that as `(settled)`. Existing installs pick this up when `install.sh` rebuilds `omacosy-helper`.

https://claude.ai/code/session_01VaidEMMJiYMGJ8QYBVm1YZ


datarip:
Thanks for the credit on #17, and for writing the better fix. All four of your
objections to my patch are right, and two of them I had not seen. `soloAndFull`
only covers closing down to **one**: close one of three and the survivor
re-expands from a quarter slot to a half and still reads stale. And a window
moved or hidden away has exactly the same shape while being alive, so no
liveness test separates them. Putting an `aerospace list-windows` on the read
path is the thing `predicted` exists to avoid. Checking after the split instead
of guarding before it is the right way round.

One small correction, not a defence: the main-display point was fixed in my
branch after I filed the issue. It resolves the display under the window's
centre with `CGGetDisplaysWithPoint` and falls back to the main display only
when that fails. The patch pasted into #17 is the older version. I mention it
only so the branch is not judged on it — the other three objections stand and
they are enough on their own.

## It has a live AeroSpace session now

You wrote that it is not verified against one, and that the fix assumes
AeroSpace starts re-expanding the survivor inside the settle grace. I can close
that for you. MacBook Air M1, macOS 27.0 (26A428), single 1440x900 built-in
panel, no notch, AeroSpace 0.21.3-Beta.

Your branch is the base of mine and I have not touched the correction path, so
what follows is your code running.

963 hints logged. **37 took the correction, and the assumption holds in all 37.**
Thirteen of them are the #17 case exactly — the survivor read at its stale half
slot, split the wrong way, then corrected:

```
1789256423.4095879 wid=1347  708x853 (read)    -> v rc=0
1789256423.8284721 wid=1347 1424x854 (re-read) -> h rc=0
```

Across all 37, the gap between the two log lines was 240 ms at the fastest,
497 ms median, 700 ms at the slowest, which includes the second `aerospace
split` where one was issued. Twenty issued a second split returning 0, nine held
their direction and issued nothing (`rc=-`).

`split` on an only child turning the container, rather than nesting, also holds
in practice: those twenty second splits are invisible on screen.

## A race: the read is stamped as fact before it is verified

```swift
let stamp = "\(wid) \(w) \(h) \(now)"
try? stamp.write(toFile: statePath, atomically: true, encoding: .utf8)
...
guard how == "read", let f = settle().frame, (f.2, f.3) != (w, h),
    (try? String(contentsOfFile: statePath, encoding: .utf8)) == stamp else { exit(0) }
```

On a `read` the possibly-stale size is written to the state file **first** and
settled afterwards. If a window opens inside that settle — up to 5 × 75 ms —
two things happen at once:

1. the newcomer's `predicted` branch chains off the stale size, and
2. the newcomer's own write to `statePath` makes the `== stamp` compare unequal,
   so the correction never runs.

The chain is poisoned exactly as #17 describes, and the guard that protects the
newcomer is what cancels the repair.

Reproduction is narrow by hand: a close, then an open inside 375 ms. Reaction
time after a deliberate `Super+W` is longer than that, so I drove it from a
script, and it reproduces every time:

```sh
aerospace close --window-id <id>
sleep 0.2                 # inside the 5 x 75ms settle
omacosy-spawn Ghostty     # the newcomer logs (predicted) at half the stale slot
```

A burst through `omacosy-spawn`, or a held key repeat, is not bound by reaction
time either.

The fix I would suggest keeps your design. Mark a read-derived state line
**provisional**, have `predicted` refuse to chain off a provisional line, and
let the correcting run rewrite it as verified. No delay before the split, no
query on the fast path, and "split fast, check after" stays intact. It also
fails towards measuring, which is the safe direction: a wrong refusal costs one
settle, a wrong acceptance poisons the chain.

## A second bug, in the branch this one leaves alone

```swift
if let s = state, wid > s.wid {
```

Ids only increase, so this was meant to mean "a window that did not exist a
moment ago". It only means "higher than the one window we happened to hint
last", which any older window with a larger id also satisfies.

Two windows side by side are enough. Clicking from one to the other and back,
every landing on the higher id took `predicted` and halved a slot that nothing
was splitting — six clicks, six mispredictions, none correct:

```
1789047228.1502361 wid=1824 708x883 (settled)   -> v rc=0
1789047229.6229281 wid=1953 708x441 (predicted) -> h rc=0
1789047230.719818  wid=1824 708x883 (read)      -> v rc=0
1789047231.365044  wid=1953 708x441 (predicted) -> h rc=0
1789047232.431158  wid=1824 708x883 (read)      -> v rc=0
1789047233.0884008 wid=1953 708x441 (predicted) -> h rc=0
```

wid=1953's real frame was 708x883 throughout — portrait, so `v`. It was hinted
`h` on every click.

There is a second half. The count alone cannot tell "one more window here" from
"one window on a different workspace" — both read as 1. I watched a terminal
opened alone on an empty workspace get predicted at 720x900, half the width of a
fullscreen window on the workspace I had just left, against a real slot of
1424x883. Portrait, so the hint said stack, and the pair ended up one above the
other.

## The branch

`fix/split-hint-unverified-read-and-prediction`, one commit, cut from **this
branch** rather than `main`, so the diff is only my delta and GitHub will
retarget it when this merges.

Both fixes are in that one commit because they need the same field on the state
line, and doing them apart means changing the format twice:

```
yours:       wid w h ts
this branch: wid w h ts maxWid count verified ws
```

`maxWid` is the highest id ever hinted, which is what the id compare was
reaching for. `count` is the focused workspace's tiled window count, and `ws` is
that workspace. `verified` is the provisional flag above. A line in the old
format still seeds `maxWid` and reads as unverified, so its prediction is
refused — which is what an unknown state should do.

It costs one `aerospace list-windows`, 21 ms, per run that writes a verified
line. It is issued after `split` and after the settle, never before the hint,
and the predicted path reuses the answer it just checked. A burst of five spawns
still chains with no extra measurement.

**Worth stating plainly: my guards do not close the race above.** After a
close-refocus stamps the stale size, a new window gives `count + 1` and a
matching workspace, so it would chain off the stale size anyway. The two are
independent and I think both are needed. This is not a competing fix for the
same thing.

## Two housekeeping notes

**My own patch is withdrawn.** `fix/split-hint-stale-frame-after-close` is not
coming as a PR. Yours is broader and cheaper and there is no point sending a
narrower version of a fix you have already written.

**One other branch of mine also collides with this one.** It changes
`splitWidthMultiplier` from a constant into a value decided per display, for
narrow screens, and you moved that declaration here. Trivial conflict, separate
fix. Say whether you would rather have it before or after this merges and I will
send it that way round.

If you would rather write the provisional flag yourself now that the
reproduction is written down, that is fine too — close mine and I will not mind.

paulsp94:
Pushed a second pass. A review of my own branch found two holes in the correction, both now closed.

**The stale slot was published as fact.** It went into the state file the moment it was read, so a window opening inside the grace chained off it and halved it. That is the damage #17 really does, and the correction arrived too late to undo it. A read is now published **provisional** (a fifth field), and becomes verified only once the watch confirms the slot. A provisional chain is never halved — that window takes the settle path instead.

**The newer-hint guard only saw newcomers that write a hint.** A window can join the container without writing one: `omacosy-layout togglesplit` retiles in place, and a window can open without taking focus. A second `split` would then nest it. The correction is **grow-only** now, which is the physical difference between the two cases: a survivor re-expanding after a close grows, a window joining shrinks the slot. A shrink still publishes the true slot, it just never re-splits. 2pt of tolerance keeps jitter from counting as movement.

Also fixed the claim about the watch: the grace is ~375ms only while nothing moves, up to ~1.2s once something does.

**Compatibility:** old four-field lines still count as verified, so an old chain is not orphaned; an *older* helper ignores the five-field line and falls back to its settle path, which is safe. `omacosy-spawn` and `omacosy-layout` read field one only — verified.

26 scenarios now run against a real `CGWindowID` whose size follows a script: grow, shrink, jitter, provisional and legacy chains, the newer-hint cancel, the predicted fast path staying under 100ms, and main's helper still reproducing #17.

@datarip thanks for the live run — 963 hints with 37 corrections is exactly the evidence the branch was missing. Note the two fixes above change what it publishes, so if you re-run it, the interesting line is now whether any `(re-read, shrank)` appears.

PR #21, opened 2026-09-13
fix(bar): the activity chip keeps the old theme's accent

datarip:
Switch theme and every pill takes the new accent except the activity
chip, which stays in the colour of whatever theme was current when the
bar started. Seen going from tokyo-night to gruvbox: the Apple logo and
the app name turned orange (ACCENT 0xfffe8019) while the chip stayed
blue (0xff7aa2f7).

It is the only right-hand item built once at startup rather than by an
updater:

    rightItems["activity"] = BarItem(icon: "󰍛", iconColor: palette.accent)

That copies the accent by value at launch. Battery, wifi, brightness,
volume and the clock all go through set() from their own updaters, so
they are rewritten on the next tick and pick the new palette up for
free. Nothing ever rewrites this one, and the theme watcher reloads the
palette and repaints without touching it.

The watcher now re-applies the accent. set() compares before it
repaints, so a theme whose accent is unchanged owes no pixels.

Reproduce: start the bar under one theme, Super+Shift+T to another with
a different ACCENT, and look at the rightmost chip.


paulsp94:
Diagnosis is right, and it is the real cause rather than a symptom: the chip is built once with `iconColor: palette.accent` (`helper/bar.swift:3599`), the draw path re-reads the live palette only when `iconColor` is nil (`:2632`), and the theme watcher (`:3414-3423`) reloads the palette and repaints without touching `rightItems`.

**One change.** Going through `set()` costs a second repaint of every surface: `set()` calls `repaint()` itself (`:612`), and since the accent has changed, the "no pixels owed" early return cannot fire — precisely the case this PR exists for. `rightItems["activity"]?.iconColor = palette.accent`, placed just above the handler's existing `repaint()`, is equally minimal and paints once. Placement before `iconCache.removeAll()` is harmless either way; that cache holds app icons keyed by name, no colours.

**One scope question.** Battery snapshots `palette.red/green/yellow` (`:651-660`) and brightness stores `palette.muted` while shaded (`:855`), so both keep the previous theme's colour in the same way, repainting only when their own publisher fires. The description says battery picks the new accent up for free, which is not the case. Either fix those two here or drop the claim.

datarip:
Thanks. Both points are answered in b11e5f1, in a slightly different shape from the one you suggested:

- **One repaint:** the chip line in the theme watcher is gone, and the watcher is back to what `main` has. `BarItem.iconColor` is a `KeyPath<Palette, NSColor>` now, looked up in the live palette when drawing (`item.iconColor.map { palette[keyPath: $0] } ?? palette.label`). The watcher's own `repaint()` recolours the chip, and `set()` never fires for it.
- **Scope:** battery (`\.red`, `\.green`, `\.yellow`, `\.label`) and shaded brightness (`\.muted`) are fixed by the same change, so they follow a theme switch in the same repaint. The description is corrected: the battery claim is gone.

The direct assignment would have fixed the chip only, and each pill that stores a palette colour would need its own line in the watcher. With the key path, a new pill cannot bring the bug back. The branch diff is 12 lines in, 10 out.

Tested on macOS 27, live: tokyo-night → gruvbox → three derived palettes. The chip took each accent in the switch's repaint, and the battery icon (label colour at 60%) followed. Not exercised live: the battery's red/yellow/green thresholds and shaded brightness; they take the same path.

PR #22, opened 2026-09-13
fix(theme): the first Super+Shift+B after a theme change does nothing

datarip:
theme-bg-next remembers the current wallpaper as a symlink and cycles to
the next image in the theme. The comparison that finds the current one is
a STRING compare, and the two sides spell the same file differently.

theme-set writes the resolved repo path:

  /path/to/omacosy/themes/tokyo-night/backgrounds/0-winding-road.webp

theme-bg-next builds its list with find under the current-theme symlink:

  ~/.config/omarchy/current/theme/backgrounds/0-winding-road.webp

Same inode, different string, so the loop never matches. NEXT keeps its
initial value, which is the FIRST image, and theme-set has just set that
image. The key appears dead. Press it again and it works, because by then
theme-bg-next has written the link in its own spelling.

There is a third spelling. The argument form, theme-bg-next <path>, writes
"$(cd "$(dirname "$1")" && pwd)/$(basename "$1")", so a wallpaper set by
hand leaves the cycle stuck in the same way.

-ef compares the inode two paths reach, so every spelling of one file
matches and no writer has to change. readlink goes with it: -ef follows
the link itself, and a missing or dangling link is false, which keeps the
same fallback to the first image the readlink guard gave.

Tested on a theme with seven backgrounds, one press per line:

  after theme-set          0-winding-road -> 1-quattro     (was: no move)
  plain cycle              1-quattro      -> 2-swirl-buck
  after the argument form  3-sunset-lake  -> 4-omakub      (was: no move)
  wrap at the last image   6-oma          -> 0-winding-road
  link deleted             any            -> 0-winding-road
  link dangling            any            -> 0-winding-road


paulsp94 (approved):
LGTM

PR #25, opened 2026-09-13
fix(config): size the top gap from the display's safe-area inset

datarip:
The shipped config gives the built-in display a 10 point top gap and
every other display 42:

    outer.top = [{ monitor.'built-in' = 10 }, 42]

The comment above it states the reasoning correctly. On a notched
display macOS reports usable space as starting below the camera strip,
so a window placed from there is already clear of the strip and needs
less reserved for the bar. The rule then matches on being a built-in
display. Built in and notched are different sets.

The notch arrived with the 14" and 16" MacBook Pro in October 2021.
Every Apple Silicon laptop before it is built in and has none: the
MacBook Air M1 and the 13" MacBook Pro M1, both 2020, along with every
Intel Mac. On those the inset is 0, so nothing is added to the 10. The
bar is 34 points tall and sits at the top of the panel below normal
windows, so tiled windows are drawn over it. The bar looks like it has
vanished: visible on an empty workspace, gone the moment anything is
tiled.

Ask the screen rather than guess from the hardware class.
NSScreen.safeAreaInsets.top is the offset macOS reports for a display,
0 where there is no notch. Apple maintains the number, so no table of
Mac models is needed and none can go stale when new hardware ships. A
model table would also answer the wrong question: the gap does not
depend on how tall the notch is, it depends on where macOS says the
usable area starts.

A new safe-top subcommand prints that offset for the built-in display.
It reuses builtinDisplayID(), already in this file for the brightness
subcommand, and matches that display to its NSScreen through the
NSScreenNumber key in deviceDescription, the same lookup bar.swift
uses. install.sh runs it while generating aerospace.toml and fills a
new @OUTER_TOP@ placeholder. The project already reads safeAreaInsets
twice for notch questions, in BarSurface.notched and in the displays
subcommand, and both reduce it to a yes or no. This is the first place
that needs the quantity.

The gap itself:

    outer.top = max(BAR_MARGIN, BAR_HEIGHT + BAR_MARGIN - inset)
              = max(8, 42 - inset)

34 is barHeight in bar.swift. 8 is the gap the other three edges and
both inner gaps already use. Their sum is 42, which is the flat-panel
value the config already ships, and 42 minus a 32 point inset is 10,
which is the built-in value it already ships. Both shipped numbers are
this calculation done by hand for two machines. Hardware that is
already correct keeps the number it has, and the machines that were
wrong become right.

Three display types, one rule:

  Notched built-in. macOS excludes the camera strip from the usable
  area, so the 34 point bar occupies space no window could take and
  only the margin is left to reserve. inset 32 gives 10, the value
  shipped today.

  Notchless built-in. macOS excludes nothing, so the bar's whole height
  is reserved as well as the margin. inset 0 gives 42.

  External monitor. Behaves as a notchless built-in and keeps the
  unconditional 42 in the list's second entry.

The per-monitor list stays for that reason. AeroSpace applies it per
monitor while it runs, so a notched MacBook docked to an external
display gets 10 on the built-in panel and 42 on the external one at the
same time. Collapsing the line to a single number would hand the
external monitor the notched value and bury the bar there.

That same difference is why the two bars rest differently per display,
which is the subject of the bar auto-hide change that follows this one.
On a notched display the strip is not usable screen. Hiding the omacosy
bar there reclaims nothing, so it stays visible, fills the strip and
covers the native menu bar, and both bars show by default. On a
notchless display and on an external monitor that strip is ordinary
screen, about 34 points of it, so the bar hides at rest and gives it
back, and the native menu bar is auto-hidden alongside it. The
behaviour follows the display rather than the machine, which is what a
notched MacBook with an external monitor attached needs.

That change is also where the gap on notchless displays and external
monitors drops to the plain 8 point margin. Once the bar hides at rest
it takes no usable space, so BAR_HEIGHT leaves the sum. Notched
displays are unaffected, because the inset already covered them. This
commit lands first and makes the number derived. The auto-hide change
then alters one input to it.

install.sh builds omacosy-helper further down the same script than the
config it generates, so a first install has no binary to ask. The gap
falls back to 42 there, and on any unusable output. Too large wastes
space, too small buries the bar, so failing large fails safe.

BAR_HEIGHT=34 in install.sh mirrors barHeight in bar.swift and nothing
enforces the pair. If the bar height changes, that constant has to
change with it.

Measured on a MacBook Air M1, macOS 27.0 (26A428), 1440x900, no notch,
AeroSpace.
Before, with the shipped gap of 10, a tiled window landed at y=10 and
covered 24 of the bar's 34 points. After, with the computed 42, it
lands at y=42 and clears the bar by the same 8 points as every other
edge. The notched case was exercised two ways: feeding the calculation
an inset of 32, which produced 10, and making macOS reserve 31 points
at the top so a gap of 10 had something to sit below. The window landed
at y=41 and cleared the bar by 7. Real hardware at inset 32 puts it at
42 and clears by 8.

One finding worth recording for anyone tempted to re-evaluate this at
runtime instead. AeroSpace caches the screen geometry when it starts,
and reload-config does not re-read it. Three measurements above read
y=10 until AeroSpace was restarted. A hook that rewrote aerospace.toml
on a display change and called reload-config would not move the gap.
Install time is where this can be decided without restarting the window
manager.

THE HEIGHT IS ASKED FOR TOO, NOT JUST THE INSET

An earlier version of this wrote `BAR_HEIGHT=34` with a comment pointing at
helper/bar.swift. That was a guess mirroring another file, and the two had
already diverged: the bar takes its height from NSMenu.menuBarHeight, and
macOS answers 30 on this machine. The gap was reserving 42 where 38 was
needed, losing four points of every window, and a hardcoded number is wrong
in whichever direction Apple next moves it.

`omacosy-helper bar-height` prints what macOS draws, the same way `safe-top`
prints the inset rather than matching a table of models. install.sh uses both
and keeps 34 only as the fallback for a first install, where the helper does
not exist yet.

macOS will not tell a process how tall the menu bar is unless it has a menu
of its own, and a command-line tool has none. An empty NSMenu is enough to be
told, and draws nothing.

---

## What it looks like before the fix

The bar appears to have vanished. It is visible on an empty workspace and gone
the moment anything is tiled. It is not hidden, it is covered — tiled windows
start 10 points from the top and the bar rests below normal windows:

```
/tmp/omacosy-borders.log
show app=zen     f=8,10   708x881
show app=ghostty f=724,10 708x437
```

Hovering the top edge does not rescue it. The bar rests at level -20 and only
climbs when `fullscreenDisplays()` reports the display as covered, which never
happens for ordinary tiled windows. So on a notchless built-in display the bar
looks broken out of the box.

## Checking the inset on your own machine

```swift
import AppKit
for s in NSScreen.screens {
    print("\(s.localizedName)\t\(Int(s.frame.width))x\(Int(s.frame.height))\ttop inset: \(s.safeAreaInsets.top)")
}
```

Here it prints `Built-in Retina Display  1440x900  top inset: 0.0`, which is the
whole bug in one line: built in, and no notch. After this change the same
machine gets 42 instead of 10.

## Why this is decided at install time, not at runtime

I looked at doing it live on a display change, and it does not work. Checked
rather than assumed:

| | |
| --- | --- |
| A command to set gaps | **No.** Nothing gap-related in `aerospace --help` |
| A way to read the current gap | **No.** `config --get` covers bindings only; `config --major-keys` returns `.` and the mode tables |
| The config file path | **Yes.** `aerospace config --config-path` |
| An event for display changes | **No.** `subscribe --all` during a real monitor plug produced only `focus-changed`, `focused-workspace-changed`, `focused-monitor-changed` and `binding-triggered` |

So a runtime fix would have to rewrite `aerospace.toml` blind, listen to AppKit
rather than to AeroSpace, and call `reload-config` — and as the commit message
says, `reload-config` does not re-read the screen geometry AeroSpace cached at
startup, so the gap would not move anyway. Install time is where this can be
decided without restarting the window manager.

The per-monitor list is kept for the docked case, so a notched MacBook with an
external display attached still gets 10 on the built-in panel and 42 on the
external one at the same time.


datarip:
Amended this branch just now, `72182d0` → `033cdfd`. Force-push, no rebase; the diff gains three lines.

**What changed.** The second entry of `outer.top` was still the literal `42`. It is now filled from a new `@OUTER_TOP_EXT@` placeholder.

**Why it belongs in this commit rather than a follow-up.** This PR adds `omacosy-helper bar-height` on the argument that a hardcoded 34 is a guess, and that macOS answers 30 on the machine it was measured on. `42` is `34 + 8` written out. Leaving it meant the built-in resolved to 38 while every other display stayed at 42 — one setting disagreeing with itself, using the number this same commit disproves.

```sh
OUTER_TOP_EXT=$(( BAR_HEIGHT + BAR_MARGIN ))
OUTER_TOP=$OUTER_TOP_EXT
```

No external display has a notch, so its gap is the inset-0 case — which is exactly where `OUTER_TOP` already started, before the built-in's inset is subtracted. The built-in's arithmetic is untouched.

**Measured with an external monitor attached.** MacBook Air M1, macOS 27.0:

| | built-in | LG UltraFine |
| --- | --- | --- |
| logical size | 1440 x 900 | 1920 x 1080 |
| `safeAreaInsets.top` | 0 | 0 |
| generated `outer.top` | 38 | 38, was 42 |

`install.sh` reads both numbers before it builds the helper, so a first install falls back to 34 and writes `42, 42`, and a second install writes `38, 38`. They agree either way, which the literal did not.

PR #26, opened 2026-09-13
fix(aerospace): Super+Shift+F can only ever open one Finder window

datarip:
`Super+Shift+F` is your own binding, in `main`'s aerospace template. It opens
**one** Finder window, once, and a second press does nothing visible.

`open <dir>` raises an existing window for that directory instead of making a
new one. That also explains the "teleport" people report: raising it activates
Finder, and AeroSpace follows the app to its workspace.

Finder cannot be opened with `open -na` either — it is always running, so
`open -na Finder` fails with "Launch failed". A new Finder window takes
AppleScript.

## Why it goes through `omacosy-spawn`

The obvious fix is to call `osascript` from the binding. That loses the lock
`omacosy-spawn` has held for `Super+Enter` since forever, and the lock is
load-bearing here.

`split-hint` runs off `on-focus-changed`, and a burst into an app that is
**already frontmost produces ONE focus change, not four**. Measured: four
`Super+Shift+F` at once ran the hint once and gave five equal 285 pt columns.
Through the lock it ran four times and split each slot.

So `omacosy-spawn` grows `--cmd <command...>`: the same serialisation for a
window that `open -na` cannot make. Only the command differs, so only the
command is a parameter.

```sh
if [ "${1:-}" = "--cmd" ]; then shift; SPAWN=("$@")
else SPAWN=(open -na "$1"); fi
```

## It collides with #12, and the fix is one line

#12 (`perf/ghostty-spawn-reuses-instance`, @jborlum) changes the same two
`open -na "$APP"` call sites, to `spawn_app "$APP"`. **His fast path is the
better default and should win.** The two are not rivals: his is a
specialisation of the default case, mine is a generalisation around it. They
compose in one line.

```sh
else
  [ -n "${1:-}" ] || exit 1
  SPAWN=(spawn_app "$1")     # instead of  (open -na "$1")
fi
```

A function name works in that position as any command does. `Super+Enter`
keeps his ~300 ms → ~140 ms, `--cmd` is untouched.

I would have stacked this on his branch so the diff showed only the delta, as
I did for #20. It is not possible here: #12 comes from `jborlum/omacosy`, so
there is no base branch in this repository to target. Merge either order and
resolve those two lines as above — or say the word and I will rebase onto
whichever lands first.

It is already running composed on my own fork, so the combination is tested,
not proposed.

## It also touches #19, trivially

`uninstall.sh` — one shared list of binaries to remove. Both branches add a
name to it, so git flags the line. I checked with `git merge-tree`: that file
is the only collision between the two, and the resolution is to keep both
names. It disappears entirely if this merges first.

## Documentation

`README.md` gains the new command and the new flag, because neither was
documented anywhere a user would look:

- the `Super+Shift+F` row in the keybinding table, and a paragraph under that
  table covering `omacosy-finder-window` and `omacosy-spawn --cmd`
- the **Automation** row in the permissions table, which now names Finder.
  That table promises to list every prompt this setup can raise, and this
  change adds one

## Cost

One Automation permission grant, for Finder, the first time the binding is
used. Finder's 474 pt minimum width is untouched and still applies: four
Finder windows do not fit on a 1440 pt display in any layout. The fix is the
SHAPE, not the fit.

## Tested

Four `Super+Shift+F` at once give four split slots rather than one flat row.
A single press gives one window, and a second press gives a second window.


paulsp94:
One line short of complete.

**Missing binding.** `bin/omacosy-karabiner-omniwm:80` is still `rule("files", "f", SHIFT, cmd="open -a Finder")`, and it is not among the six files here. Under OmniWM the chord keeps the old behaviour — and `open -a Finder` only activates, which is the worse half of the bug — while bypassing the spawn lock and the preselect entirely.

**Then the README follows.** `:321` and `:333` promise a new window on every press, which is not true under OmniWM until that line lands.

The rest is in good shape. `bin/omacosy-finder-window` is 62 lines of which 57 are comment, so the real code is five lines plus about twelve in `omacosy-spawn`; `--cmd` generalises cleanly; music and messenger are untouched, so nothing spreads silently; and the Automation grant the osascript needs is recorded in the README permissions ledger, with install and uninstall registration correct.

**One note on the title.** With no window targeting `~` at all — none open, or navigated elsewhere — `open ~` does create one. "Can only ever open one" is the steady state rather than an absolute.

datarip:
Thanks. Rebased onto `main` (the conflict was `omacosy-spawn`: `--cmd` now sits beside your `spawn_app`, so Super+Enter keeps the Ghostty fast path), and 13a0c45 answers the review:

- **Missing binding:** `bin/omacosy-karabiner-omniwm` runs `omacosy-finder-window` for `files` now, so the chord takes the spawn lock and the preselect under OmniWM too. Checked with a temporary HOME: the generated rule calls it, and `remove` still takes all 34 rules out.
- **README `:321` / `:333`:** true under both managers with that line.
- **Title:** changed to "raises the Finder window instead of opening a new one", and the README paragraph now says `open ~` makes a window only when none shows the home folder.

Tested under OmniWM on an empty workspace: three presses 1.5 s apart give three Finder windows in the dwindle staircase (708 × 854, then 708 × 423 twice), the same shape as three Super+Enter presses.

datarip:
Heads-up on merge order: #26 and #44 both edit `bin/omacosy-spawn` and conflict with each other, in either order (checked with `git merge-tree` against current `main`). #28 merges cleanly with both.

Whichever of #26 and #44 you merge second, tell me and I will rebase it. The resolution is #26's `--cmd` wrapped around `spawn_app`, together with #44's lock loop; my fork has run that combination daily.

paulsp94:
Needs a rebase

datarip:
Rebased onto `main` (`3d66f9d`). The new commits are `70f3b51` and `50f598a`.

There was one conflict, in `bin/omacosy-spawn`. #44's wall-clock wait meets this branch's `--cmd` on the AeroSpace path. I kept `main`'s `until [ "$SECONDS" -ge ... ]` loop and changed only the call before it, from `spawn_app "$APP"` to `"${SPAWN[@]}"`. The OmniWM path took the same call with no conflict. `git range-diff` shows no other change to the two commits.

PR #28, opened 2026-09-13
fix(tiling): a read is a guess until it settles, and a prediction needs a live layout

datarip:
**Your branch is the base, and it is the better fix.** `fix/split-hint-stale-read`
splits fast and checks after, so it costs nothing until a correction is due, and
it covers close, move, hide and 3 → 2. My own `fix/split-hint-stale-frame-after-close`
is withdrawn: it blocked 50 ms and spawned an `aerospace list-windows` on every
refocus, and its `guard count == 1` only ever covered closing down to one window.

Two holes are left. They are in one pull request because both need the same field
on the state line, and doing them apart means changing the format twice.

## 1. A race: the unverified read is stamped as fact

On a `read` the possibly-stale size is written to the state file first and settled
afterwards. Open a window inside that settle — up to 5 × 75 ms — and two things
happen at once:

1. the newcomer's `predicted` branch chains off the stale size, and
2. its own write to `statePath` makes the `== stamp` compare unequal, so the
   correction never runs.

The chain is poisoned exactly as #17 describes, and the guard that protects the
newcomer is what cancels the repair.

**Reproduction.** A human cannot hit a 375 ms window, so it is scripted:

```sh
aerospace close --window-id <id>
sleep 0.2                 # inside the 5 x 75ms settle
omacosy-spawn Ghostty
```

Before: the newcomer logs `(predicted)` at half the stale slot. After: `(settled)`
at its real one.

**The fix** is one field, `verified`. A `read` writes `0`; the settle writes `1`
with the settled size; `predicted` refuses a line without it. Your "split fast,
check after" survives intact — no delay, no query on the fast path — and it fails
towards measuring, which is the safe direction: a wrong refusal costs ~375 ms
once, a wrong acceptance poisons the chain.

## 2. `predicted` chains off a layout that no longer exists

Separate bug, in a branch your fix does not touch. `predicted` fires when the
focused window's id beats the id in the state file. Ids only increase, so the
comparison was meant to mean "a window that did not exist a moment ago". It only
means "higher than the one window we hinted last", which any older window with a
larger id also satisfies.

Two windows side by side are enough. Clicking from one to the other and back
mispredicted on every click onto the higher id, 12 out of 12, each choosing the
opposite direction to the frame the window actually had.

The 3 s TTL bounds staleness in time, not in layout. Three fields bound it in
layout: `maxWid` (highest id ever hinted, so "new" is actually testable), `count`
(the focused workspace's tiled window count — a close and an open net back to the
stored value and are refused) and `ws` (the count alone cannot tell "one more
window here" from "one window somewhere else": both read as 1).

## The line

```
yours:       wid w h ts
this branch: wid w h ts maxWid count verified ws
```

`verified` goes before `ws`, because `ws` is a name and is parsed as the last
field. A line in the old format still seeds `maxWid` and reads as unverified, so
its prediction is refused, which is what an unknown state should do.

## What it costs

One `aerospace list-windows`, 21 ms, per run that writes a verified line. It is
issued after `split` and after the settle, never before the hint, and the
predicted path reuses the answer it just checked.

A read that turns out to be right now writes `verified = 1` and logs
`(confirmed)`. Without that, every hover refocus would leave a line nothing can
chain from, and the burst after a hover would measure for nothing.

## Tested

Driven from a script, against `/tmp/omacosy-split-hint.log` rather than the
screen: 12 refocuses with no misprediction, a lone window on an empty workspace
measured rather than chained, five spawns in a row still chaining, the 375 ms race
above, closing one of three then opening, and a state file in the old format.


paulsp94:
Heads up: this went conflicting through no fault of yours. It is based on my #20 branch, and I force-pushed that on Monday after a review found two holes in it — a read is now published *provisional* until the watch confirms it, so a window opening inside the grace can no longer chain off a stale slot, and the correction is grow-only, since a window joining the container shrinks it while a survivor re-expanding grows. Worth rebasing onto the new head before we take this further; some of what you were closing may already be closed.

Two notes from reviewing this branch against it:

- `wid > maxWid` is a real improvement and free — the field is already on the line, and `wid > s.wid` genuinely mispredicts against any pre-existing window with a higher id.
- But `chainable` short-circuits only when `wid > maxWid`, which is exactly a genuine burst spawn, so every real prediction now runs `aerospace list-windows`. That is the path `predicted` exists to keep free of subprocesses. If the workspace is the part you need, `/tmp/omacosy-bar-ws` already holds the focused workspace, written by the exec-on-workspace-change hook.

Also worth a look: `verified` currently means "did not move within the 375 ms grace" rather than "measured", so if a re-expansion outlasts the grace it stamps the stale size as verified.

paulsp94:
#20 is merged, so this can come off my branch now: **retarget the base to `main` and rebase.** It still points at `fix/split-hint-stale-read`, which is why it shows as conflicting.

The rebase has one real snag worth planning for rather than resolving line by line: **the two state-line formats disagree at the fifth field.**

- `main` now writes `wid w h ts verified` — five fields, where field five is `1`, or `0` while a read is still unchecked — and parses with `f.count >= 4`, taking field five as the flag (`helper/main.swift:445-446`, `:471-472`).
- This branch writes `wid w h ts maxWid count verified ws`, so field five is `maxWid`.

Read by `main`'s parser, a line from this branch turns `maxWid` — a large window id, never zero — into the verified flag, so every chain reads as verified, which is exactly the poisoning #17 is about. The other way round, this branch rejects `main`'s five-field line and loses the chain. Both can happen for real during an update, when one helper writes and the other reads.

The clean fix is to **build on `main`'s five fields and append yours after them**: `wid w h ts verified maxWid count ws`. An older reader then still parses the first five correctly, and `omacosy-spawn` and `omacosy-layout` only ever read field one.

That also shrinks this PR, because `main` now does part of what it set out to do: a read is published provisional until the watch confirms it, a provisional chain is never halved, and the correction is grow-only. So `verified` can be `main`'s rather than a second definition of it.

What I'd still keep from here is `wid > maxWid`, a real improvement, since `wid > s.wid` mispredicts against any pre-existing window with a higher id. What I'd still drop is the `aerospace list-windows` inside `chainable` (and `highestKnownWid()`), which puts a subprocess on the predicted path on every genuine burst spawn — the one path built to have none. The focused workspace is already in `/tmp/omacosy-bar-ws` if that's the part you need.

Happy to review again once it's rebased.

datarip:
Thanks. Retargeted to `main` and rebuilt on it as 4ab3945 (63 lines in, 12 out, down from 179); the description is rewritten.

- **The field-five clash:** the line is `wid w h ts verified maxWid ws`, with `main`'s five fields first, and `verified` is `main`'s own. Measured with `main`'s helper: it reads this line correctly, and it misreads the old eight-field one, as you said.
- **What `main` already closed:** the provisional read and the grow-only correction. My own `verified` definition is gone.
- **`wid > maxWid`:** kept, with a seed from the window server's own list when there is no state, in process.
- **`aerospace list-windows` / `highestKnownWid()`:** both gone. The workspace comes from `/tmp/omacosy-bar-ws`, as you suggested. It is still needed: `main` failed "lone window after a fullscreen workspace" 5 times out of 5 in the benchmark. The window count is dropped; your provisional read covers the close-then-open case it guarded.

One addition you did not ask for: `omacosy-spawn` waits while the state line is provisional. Close-then-open within 0.1 s failed for every helper, `main` included, because the new window joins the survivor's container before the stale read can be corrected, and the grow-only rule then rightly refuses a second split. With the wait it passed 5/5, and bursts are unchanged. The mechanism, a log excerpt and the full benchmark table are in the description. If you would rather have the spawn wait as its own PR, I can split it out.

Also noted in the description: `main`'s parser reads a state line with a trailing newline as the four-field (verified) format. No writer adds one today, so it cannot happen now.

Note: `bin/omacosy-spawn` is also touched by #26 and #44, so whichever of the three lands later will need a small rebase. I can do that one.

datarip:
Correction to the last note in my reply: I have now checked it with `git merge-tree`, and #28 merges cleanly with #26 and with #44, in either order. Its `omacosy-spawn` change is in a part of the file the others do not touch. Only #26 and #44 conflict with each other, and I have said so on both.

Issue #15, opened 2026-09-08
Bug: Omacosy menu bar overlaps with apple menu bar. Suggestion: dual menu bar (see video + patch)


datarip:
Hardware: Macbook Air M1 (2020 - no notch)
OS: MacOS 27 (Golden Gate Beta)
Config (Workspace/Tiling): Aerospace

---
Video:

https://github.com/user-attachments/assets/74727e41-a8de-4b1c-87f8-9bfce172f33a

---

Hi @paulsp94 and thank you for this nice little project, really cool. While most of the idea credit goes to Vaxry/Hyprland, I think what you have done can be a thing on its own, the Mac OS way ;-) BTW I found it works better with Aerospace than OmniWM in my case. More opinionated and simpler, imo.

So here below is what I've done. Yes, the report is in part AI generated, but it summarizes the issues well in addition to the video and the code diff I added in my patch below.

## The problem

With the native macOS menu bar set to auto-hide, both bars want the same
gesture. Moving the pointer to the top edge summons the native menu bar,
and the omacosy bar is already there and overlaps on top of it. Because the
bar is transparent, native menu titles show through the gaps between
pills, and the omacosy bar blocks clicks on them. Some popular Mac apps like Dropshare or Supercharge rely heavily on the menu bar, so for me, it was painful to use them like that.

Menu-bar apps are the sharp edge of this. Anything living in the native
status area becomes hard to reach while the omacosy bar covers it.

`macos-defaults.sh` already sets the native menu bar to auto-hide, so
this is the shipped configuration rather than something unusual.

## Proposed behaviour

Auto-hide the omacosy bar as well, and split the top edge by side.

| Action | Result |
| --- | --- |
| Pointer to the top-**left** edge | The omacosy bar reveals, with an opaque background |
| Pointer to the top-**right** edge | Only the native menu bar; omacosy stays hidden |
| Slide sideways along a revealed bar | It stays; crossing the middle does not swap bars |
| Pointer down, out of the strip | Both drop |
| Return to the top edge | The side you enter on decides again |

No modifier to hold. The side is latched **once** on entry to the strip
and held until the pointer leaves, which is what lets you slide along a
revealed bar to reach its right-hand pills without it dropping.

Existing behaviour is preserved: `openPopup == nil` still keeps the bar
up once a popup is open, and macOS keeps its own menu open once a title
is clicked.

## The change

Six edits, all in `bar.swift`, in `pointerAtScreenTop`,
[...]

PR #29, opened 2026-09-13
feat(bar): split the top edge between this bar and the native one

datarip:
Answers #15. **You have not replied to that issue, so this arrives uninvited** —
if the answer is no, say so and I will close it. Nothing else I have sent
depends on it, and it costs you nothing but this paragraph.

## The problem

`macos-defaults.sh` auto-hides the native menu bar so this bar can have the
strip, and the comment there states the intent: *"One bar, not two"*.

Auto-hide is not permanent hide. The native bar returns whenever the pointer
reaches the top edge — into the same strip this bar occupies, and this bar is
transparent and in the way. Apps whose only interface is a status item become
unreachable: a screenshot tool, a VPN client, a clipboard manager. The comment
above `SLSMenuBarUseBlurredAppearance` already describes what it looks like,
*"text soup over whatever sits in the strip"*.

## What it does

Split the edge. The left half belongs to this bar, the right half is left to the
native one. What that means differs by display, because the resting state does:

| display | at rest | top left | top right |
| --- | --- | --- | --- |
| notchless panel, external monitor | bar hidden, windows get the height | this bar rises | native bar arrives alone |
| notched | bar visible, as today | deliberate no-op | this bar stands aside |

A notched display has nothing to reclaim — macOS already excludes the camera
strip from the usable area — so only the right half acts there.

The side is latched on entering the strip and held until the pointer leaves, so
sliding along the bar to reach its right-hand pills does not hand the strip away
halfway across.

## Three things this costs, said up front

**1. The bar now needs Screen Recording**, which `omacosy-overview` already
asks for. It samples the native menu bar's colour once per wallpaper so an
auto-hiding bar can paint the right colour on its first frame instead of
resolving a blur on every reveal. Refused, it falls back to a live blur, which
is slower to appear and matches less exactly. `README.md`'s permission table is
updated.

**2. I cannot test a notched display.** This machine has no notch. The notched
column of that table was reached by forcing the same code path on a flat panel,
which is the path a notched Mac takes by default — so the logic is exercised,
but the camera strip itself and whether a right-half hover really lets the
native bar through need someone with the hardware. The commit message says this
again where it matters, under **A NOTE FOR WHOEVER HAS THE HARDWARE**.

**3. Two fitted constants**, `menuBarSaturation = 1.3` and
`menuBarDarken = 0.923`. They approximate the menu bar from the wallpaper
average and are the only fitted values in the file. They affect only the half
second before the first capture of a wallpaper the bar has never shown.
Leave-one-out validation over 15 wallpapers says no model of the average can do
better than about 15/255, because the average is the wrong feature. I am happy
to drop them for a plainer placeholder if you would rather the file had nothing
fitted in it.

## It is configurable, and it degrades

`config/bar.conf`, new, live copy at `~/.config/omacosy/bar.conf`:

```
autohide = auto | on | off     # auto reads the display: notched keeps the bar
split    = 0.0 .. 1.0          # where the edge divides, 0.5 to start
slide    = milliseconds | off  # the bar drops into the strip and climbs out
```

No file, or `autohide=off`, and the bar behaves as it does today. A notched
display sees no change by default.

## It conflicts with two of my own PRs, and both resolutions are one line

**#21** — `helper/bar.swift`, the theme watcher. Both add to the same callback.
Keep both: #21's `set("activity") { $0.iconColor = palette.accent }` belongs
beside this one's palette reload. If #21 merges first I will rebase this onto
the new `main`.

**#31** — `uninstall.sh`, the shared list of binaries to remove. Keep both
names.

Neither conflicts with `main`, nor with #12, #19 or #20.

## Why it is 1169 lines

Most of it is not the split. It is the colour path that makes an auto-hiding bar
possible: a bar that arrives into an empty strip has to know what colour to be
before it gets there, and four separate faults in getting that right are fixed
here, each documented in the commit message. It now reads 2.4 to 3.0 per 255
against the native bar, cold cache, with the wallpaper hammered through a whole
theme while the pointer sat on the bar.

The commit message carries the full engineering detail — the capture path, the
slide timing, the strip cache version, and what `autohide=off` actually looks
like. It is long because this is the largest thing I have sent and I would
rather the reasoning were reviewable than trusted.


datarip:
New commit: the bar stays visible by default (`autohide=off` in `config/bar.conf`).

With `auto`, every display without a notch hides the bar at rest, and that is most Macs: all non-notched laptops and every external monitor. But the top gap `install.sh` writes assumes a visible bar, so a first install on such a Mac hid the bar over an empty strip of the gap's height, until the user ran `omacosy-bar-autohide`. With `off`, the default and the gap agree from the first install. `auto` and `on` stay available.

A Mac that already has `~/.config/omacosy/bar.conf` keeps its value: `install.sh` copies the file only when none exists (`cp -n`).

datarip:
729279d fixes a case with more than one display and `autohide=off`. A top-edge hover on one display can bring the native menu bar down on another display too. Only the bar under the pointer stood aside, so the other bar stayed in front of the native one. Now each bar checks its own display.

Tested on a MacBook Air with one external display, and on a Mac mini with two external displays.

paulsp94:
Thanks for 729279d and for testing across both multi-display setups. The per-display check is right. Three things still hold this back, plus a rebase:

- **The default lives only in the shipped config.** `config/bar.conf` says `autohide=off`, but the code still falls back to `barConf.autohide ?? !notched` (`helper/bar.swift:3398`). A missing file, `auto`, or any value that fails to parse (`:3314` maps them to nil) turns autohide on for every flat panel. Please make the code fallback `false` so the config and the code agree. The comments at `:3493-3495` and `:3519-3520` still describe the old `!notched` default.
- **Opted-out bars still change behaviour.** With `autohide=off`, `:4170` raises the window to `barRevealLevel` on every show, so the bar sits at level 1002 permanently, above the native menu bar. And `askNativeBar` (`:3833-3841`) polls at 0.08 s / 0.25 s for every display (`:4016-4019`). Someone who never turns autohide on should get today's bar unchanged: same window level, and no polling.
- **The colour sampling runs without autohide too.** The wallpaper watcher calls `captureOwnStrip` for every surface (`:4617`), `captureOwnStrip` checks only `atTopEdge` (`:691`), and `seedStripFromWallpaper` runs for every surface at init (`:3529`). Please gate all of it on `surface.autohide`, as the hover capture already is (`:3949`).
- **Rebase:** it conflicts with main after today's merges (#21 and #47 touched `bar.swift`).

Smaller: the PR body's table still calls the notched top-left a "deliberate no-op", but for non-autohide surfaces the yield now ignores the split (`:4016`, commented at `:4003-4008`). One of the two should change so they agree.

datarip:
Thanks. You were right about the level and the polling, and testing your proposal showed me a simpler way to keep what the raise was for. Done in 064ea99, on top of a rebase onto `main`:

- **Default.** A missing `bar.conf` or a value the bar cannot read now means `off`, in the code as in the config. `auto` still works when it is written. The two comments that called `!notched` the default are corrected; the second one was a leftover and is gone.
- **Level and polling.** With `off`, the bar no longer rises to 1002, and the stand-aside is gone: `askNativeBar`, `setYielded` and `SLSIsMenuBarVisibleOnSpace`. Nothing in the bar polls now. The bar rests at level 20 instead of -20. The reason: with your proposal exactly as written, on my MacBook Air, the bar was invisible on a normal screen. An app kept an invisible window over the whole display at layer 3:

  ```
  layer 3     LanguageTool for Desktop   0,0 1440x900
  ```

  At 20 the bar is above windows like that and below the native menu bar (24), so the native bar comes down over it by itself. That is what the stand-aside was imitating. The window level differs from `main`'s -20, so I am telling you rather than slipping it in.
- **Colour sampling.** The startup seed, the wallpaper watcher and `captureOwnStrip` check `surface.autohide`. With `off`, the bar never paints the strip and now samples nothing.
- **Native fullscreen.** New, and related: in a native fullscreen space (`SLSSpaceGetType` 4) the bar hides in every mode and does not come back on a top-edge hover. Only the native black menu bar shows, as on any Mac. `kickVisibility` also runs at the first window event now, then at most every 0.1 s, and a space change runs it at once. Before that, a stream of resize events from the fullscreen animation kept pushing the check back, and the bar showed on the fullscreen space for 0.44 s. Measured after, in my fork's build of the same change: it hides 0.08 s before the space switch.
- **OmniWM exemption removed from `fullscreenDisplays()`.** Its comment says OmniWM 0.6.3 did not apply the outer gaps, so every tile read as fullscreen. OmniWM 0.7.2 applies them, and my fork has run without the exemption since 17 September. It matters for the native fullscreen fix above: with the exemption, under OmniWM the bar can only hide at the space switch, and I saw it flash over the window each time, in two runs of this branch's build. Without it, the bar hides as soon as the window covers the display, before the space switch, and the flash is gone. The same function now ignores a window that only touches the display edge: OmniWM parks hidden workspaces at full size one point onto an edge (1439,0, or -1439,0 with a second display to the right), and such a window used to count as fullscreen.
- **Description.** The table has one row per `autohide` value, plus native fullscreen. "Deliberate no-op" is gone with the stand-aside.

Tested on a MacBook Air M1 (no notch) and an LG UltraFine, under OmniWM, with this branch's build: `off` by hand (no flash, visible at rest, the native menu bar covers it cleanly, only the native menu bar in native fullscreen), and `on` by a script (the left-half reveal, the return and the right half 10 cycles each, native fullscreen once, all passing). Under AeroSpace I tested the same changes in my fork's build, with the same result.

One favour, if you have a moment: I only have a MacBook Air without a notch, so I cannot see how this looks beside a camera strip. Under `auto`, a notched display takes the `off` path, so the level-20 bar applies there too. Could you try these on your MacBook Pro and tell me what you see?

1. At rest, with `autohide=off`: is the bar visible, and does it sit right beside the notch?
2. Push the pointer to the top edge: does the native menu bar come down over the bar cleanly?
3. Put a window in native fullscreen and push the pointer to the top edge: do you see only the native menu bar?

If anything looks wrong, a screenshot would help me a lot. Thanks!

PR #30, opened 2026-09-13
feat(bar): one command for autohide and the top gap it implies

datarip:
> **Depends on #29 and is not meant to be read alone.**
>
> This branch is cut from #29's, because it reads `config/bar.conf`, which only
> #29 introduces. GitHub therefore shows **two commits** here: #29's, and this
> one. **Only the top commit is this pull request** — `feat(bar): one command
> for autohide and the top gap it implies`, 5 files, +188/-1.
>
> I could not stack it properly. A PR base has to be a branch in this
> repository, and #29's branch lives in my fork. If #29 merges, this collapses
> to its own commit by itself. If #29 is not wanted, close this too — it
> configures a thing that would not exist.

---

`autohide` already chose whether this bar hides at rest. Nothing made the
top gap follow it, so the two could disagree: a bar left visible with the
hiding gap has tiled windows underneath it, and a bar that hides with the
visible gap wastes its height on nothing. The only thing reconciling them
was a personal script outside this repo.

`omacosy-bar-autohide on|off|auto|status` writes the setting, resizes
aerospace's outer.top to match, reloads the config and restarts the bar.
bar.conf now says so, because editing it by hand still sets only half.

    hiding at rest    outer.top = margin
    visible at rest   outer.top = bar height + margin - notch inset

BOTH NUMBERS ARE ASKED FOR

The bar's height comes from macOS (`omacosy-helper bar-height`,
NSMenu.menuBarHeight) and the inset from the display (`safe-top`,
NSScreen.safeAreaInsets.top). Neither is a constant and neither is matched
against a table of models, so a Mac with a menu bar of a different height,
or a notch of a different depth, gets the right gap without anyone editing
a number.

The notch test is the SAME one the bar makes — `safeAreaInsets.top > 0`,
which is `notched` in bar.swift and what `autohide: barConf.autohide ??
!notched` resolves against. Asking the same API the same way is what stops
the command and the bar disagreeing about what a display is.

THE GAP IS PER DISPLAY, BECAUSE THE BAR IS

aerospace's per-monitor form is kept rather than flattened to one number.
An external monitor is never notched, so `auto` hides the bar there while a
notched built-in keeps it, and the two want different gaps. Simulated at an
inset of 74: with `off`, the built-in wants 8 and the external 38 — the
inset absorbs the bar on the notched panel, where it occupies space no
window could use anyway, and only the display that really pays 30 points
reserves them. Writing one number would have charged the notched display
for a bar that costs it nothing.

Falls back to 34 and 0 without the helper, which cannot happen on a real
install: install.sh builds the helper before it links this.

It also restarts omacosy-solo, which reads the same `autohide` key to decide
whether to fullscreen with or without the outer gaps. Without that it keeps
whatever it read when it last started: switch the bar to always-visible and
solo windows go on covering it edge to edge, so the bar hides under them and
the change looks like it did nothing. Measured after three hours of uptime,
which is exactly how long it had been wrong for.

`launchctl kickstart` on a label that does not exist fails quietly, so this
costs nothing where the solo daemon is not installed.

Documented in the README, under "The bar": the four forms of the command,
what auto reads off the display, and why this is a command rather than a
line to edit. A new user-facing command that only documents itself in its
own file header is a command nobody finds. The wording follows the
script's header and bar.conf's comments so the three cannot drift.

The solo daemon restart the command also does is left out deliberately.
It belongs to the solo-fullscreen change, it fails quietly where that is
not installed, and naming it here would raise a question this README
cannot answer.

## Conflicts

Everything below is one line in a list of installed or removed binaries, except
the first, which is inherited from #29.

| with | file | resolution |
| --- | --- | --- |
| **#21** | `helper/bar.swift` | #29's conflict, not this one's. Keep both lines in the theme watcher |
| **#26** | `uninstall.sh` | keep both names |
| **#31** | `install.sh` | keep both lines |
| **#19**, yours | `install.sh` | keep both lines |

Nothing conflicts with `main` or with #12 or #20.


datarip:
One note on scope, so it is on the record rather than a surprise later.

This covers AeroSpace. Under OmniWM the top gap comes from a separate constant — `[gaps.outer] top` in `config/omniwm/settings.toml` — and nothing in the repo writes it, so `bar.conf` has no effect there and a hidden bar still costs the whole strip.

Measured on OmniWM 0.6.10, macOS 27.0, a notchless 1440x900 built-in:

| `bar.conf` | aerospace gap | OmniWM `outerGapTop` | window top |
| --- | --- | --- | --- |
| `off` | 38 | 42 | 42 |
| `on` | 8 | 42 | 42 |
| `auto` | 8 | 42 | 42 |

The AeroSpace number moves. The OmniWM one never does, and the window never moves.

The follow-up is written and tested on my fork, but it needs `OUTER_TOP_EXT` from #25 to seed the generated settings file. So it is cleaner as its own PR once this and #25 have landed than folded in here, which would couple this PR to #25 for no benefit.

Nothing to change in this PR — just flagging that OmniWM needs a second half so it does not ship as a half-working setting for OmniWM users.

datarip:
New commit: the bar stays visible by default (`autohide=off` in `config/bar.conf`).

With `auto`, every display without a notch hides the bar at rest, and that is most Macs: all non-notched laptops and every external monitor. But the top gap `install.sh` writes assumes a visible bar, so a first install on such a Mac hid the bar over an empty strip of the gap's height, until the user ran `omacosy-bar-autohide`. With `off`, the default and the gap agree from the first install. `auto` and `on` stay available.

A Mac that already has `~/.config/omacosy/bar.conf` keeps its value: `install.sh` copies the file only when none exists (`cp -n`).

PR #33, opened 2026-09-13
fix(gesture): the OmniWM swipe config points at another user's home

datarip:
## The problem

`config/gesture/config.omniwm.json` hardcodes an absolute path under another user's home directory:

```json
"swipe_up":   "/Users/paulspende/.local/bin/omacosy-overview >/dev/null 2>&1",
"swipe_down": "/Users/paulspende/.local/bin/omacosy-overview close >/dev/null 2>&1"
```

`omacosy-wm-switch omniwm` copies that file over `~/.config/omacosy/gesture.json`, so after switching to OmniWM the 4-finger vertical swipes run a binary that does not exist and the workspace overview never opens. It works only on a machine whose username is `paulspende`.

## The fix

The sibling `config/gesture/config.json` already uses `$HOME`, and these commands run through `popen()` — so the shell expands it. This changes the two paths to match, and nothing else.

## How it showed up

macOS 27.0, OmniWM 0.6.10. After `omacosy-wm-switch omniwm`, swipe-up did nothing. `/tmp/omacosy-gesture.log` showed the daemon had loaded:

```
Loaded config: ... swipe_up='/Users/paulspende/.local/bin/omacosy-overview >/dev/null 2>&1'
```

With `$HOME` in place, the same swipe opens the overview. Note that the copy itself also needed #34 before a swapped config reached the running daemon at all — the two faults masked each other. The overview itself is fine under OmniWM — `helper/overview.swift` has an `omniwmctl` path and does not depend on AeroSpace.

## One thing left alone on purpose

`swipe_left` and `swipe_right` in the same file are `"next"` and `"prev"`. The comment above the copy in `omacosy-wm-switch` says they should be `""` under OmniWM so they cannot fight its native horizontal swipes:

> Its horizontal actions are set to "" (non-next/prev strings no-op before any aerospace call), so it cannot fight OmniWM's workspace swipes.

**Do not make the file match that comment.** Measured after opening this PR: `settings.toml` ships `[gestures] workspaceSwipeEnabled = true` with `workspaceSwipeFingerCount = 4`, so OmniWM's own handler and omacosy-gesture are both bound to the same 4-finger swipe. OmniWM's fires on any horizontal component, is not axis-committed and knows nothing about the overview overlay, so a diagonal swipe-up switches workspace *and* opens the overview — 17 out of 17 tries on my machine.

Setting these to `""` would leave OmniWM's untunable handler in sole charge and keep that bug. The better fix is the opposite: turn `workspaceSwipeEnabled` off and let omacosy-gesture own both axes, since it commits to one axis per gesture and already consumes horizontal flicks while the overlay is up. That is a separate change to `settings.toml` and not in this diff.

So this PR is only the two paths.


datarip:
Amended the description above after finding the cause of a related bug.

It previously offered, as a follow-up, to blank `swipe_left`/`swipe_right` for OmniWM so they match the comment in `omacosy-wm-switch`. That would have been wrong: `settings.toml` also enables OmniWM's own four-finger workspace swipe, so blanking ours leaves its untunable, overlay-unaware handler in sole charge — which is the bug in #35. The fix is the other way round, and #35 has the measurements.

The diff here is unchanged: just the two hardcoded paths.

paulsp94 (approved):
LGTM

PR #34, opened 2026-09-13
fix(wm-switch): a config swap reaches the daemon that reads it

datarip:
## The problem

Both handover paths in `omacosy-wm-switch` copy a new `gesture.json` into place and then call `launchctl load`:

```sh
cp "$REPO/config/gesture/config.json" "$HOME/.config/omacosy/gesture.json"
launchctl load "$HOME/Library/LaunchAgents/com.omacosy.gesture.plist"
```

`load` bootstraps a service that was booted out, but on one that is **already loaded** it fails with `Load failed: 5: Input/output error` and leaves the running process untouched. `omacosy-gesture` reads its config once, at startup, so the file just copied is never read and the daemon keeps the previous manager's swipe commands in memory.

## Measured

On this machine, switching to OmniWM and back:

| time | event |
| --- | --- |
| 16:12:48 | daemon starts, logs the OmniWM swipe commands |
| 16:12:55 | `gesture.json` rewritten with the AeroSpace ones |
| — | daemon goes on using the 16:12:48 values until restarted by hand |

`launchctl list` showed the same pid before and after the `load`.

## Why both calls are needed

`kickstart -k` cannot simply replace `load`. All four combinations, tested:

| | already loaded | booted out |
| --- | --- | --- |
| `launchctl load` | fails, `Input/output error`, pid unchanged | works |
| `launchctl kickstart -k` | works, pid changes | fails, `Could not find service "com.omacosy.gesture" in domain for user gui: 501` |

The booted-out state is exactly what `stop_aerospace_world` leaves behind, so dropping `load` would break the OmniWM path. Both, in this order, cover both states.

`ffm` keeps a plain `load`: no config of its own is rewritten here, so it has nothing to re-read.

## Scope

This is the delivery mechanism only. The visible symptom on my machine was swipe-up opening nothing, because the config the daemon held pointed at a path that does not exist — that is a separate one-line fix, sent as #33.


paulsp94 (approved):
LGTM

PR #35, opened 2026-09-13
fix(omniwm): one engine owns the four-finger swipe

datarip:
## The symptom

A four-finger swipe **up** to open the overview also switches workspace, if the swipe has any rightward or leftward component. The workspace pill moves to an empty neighbour and moves back when the overview closes — so it reads as a rendering glitch, but it is a real switch.

## Two engines, one gesture

| binding | file |
| --- | --- |
| `workspaceSwipeEnabled = true`, `workspaceSwipeFingerCount = 4`, axis horizontal | `config/omniwm/settings.toml` |
| `swipe_left = "next"`, `swipe_right = "prev"`, `fingers = 4` | `config/gesture/config.omniwm.json` |

Both listen to the same gesture. OmniWM's handler fires on the horizontal component of a diagonal; omacosy-gesture's vertical handler opens the overview; the overview restores the original workspace on dismiss, which hides the cause.

## Measured

OmniWM 0.6.10, macOS 27.0. Capturing `omniwmctl watch workspace-bar` across 17 consecutive swipe-ups gave a clean alternation:

```
event 0: focused=['3']  non-empty=['1','3']
event 1: focused=['4']  non-empty=['1','3']   ← overview opens
event 2: focused=['3']  non-empty=['1','3']   ← overview closes
...
```

The focused workspace was always one that held **no windows**, and `omacosy-gesture` logged **zero** workspace switches of its own — so the switch came from OmniWM's handler, not ours.

## Why omacosy-gesture should own it

Not preference — it has three properties OmniWM's handler does not:

- **One axis per gesture.** `fire_gesture` sets `GS_COMMITTED`, so a diagonal does one thing.
- **Overlay-aware.** `overlay_active()` consumes horizontal flicks while the overview is up. That guard exists for this exact failure, found earlier under AeroSpace: *"finger lift-off after the vertical swipe reads as a horizontal flick and switched workspaces UNDER the overlay"*.
- **Tunable thresholds.** OmniWM's commits late and exposes no knobs.

Turning OmniWM's swipe off makes the two window managers behave the same **by construction** — it is then literally the same code path under both.

## Also in this commit

The comment in `bin/omacosy-wm-switch` is corrected, because it described the opposite arrangement:

> Its horizontal actions are set to "" ... so it cannot fight OmniWM's workspace swipes.

They never were `""`. And with OmniWM's handler off they must not be, or no horizontal swipe would remain. Leaving that text would point the next reader at a change that breaks the feature — so it goes with the fix rather than after it.

## Verification

The setting reloads live, without restarting OmniWM. A diagonal swipe-up now opens the overview and leaves the workspace alone; horizontal swipes still step workspaces through omacosy-gesture.

Tested on my fork against the same OmniWM version. Not tested from a clean `main` install — the value is read by OmniWM itself, so the branch it arrives on does not affect the result.


paulsp94:
The double-handling is real — `config/omniwm/settings.toml:81-82` enables OmniWM's own four-finger workspace swipe while `config/gesture/config.omniwm.json:8-9` binds ours to the same gesture — and putting omacosy-gesture in sole charge is the right call, for the reasons this repo already gives: tuned mid-gesture thresholds, overlay awareness, cursor-monitor targeting. Keeping the now-inert keys rather than deleting them is also correct, since the decoder wants a complete table.

**The stated reason is wrong, though.** The new comment at `bin/omacosy-wm-switch:95-97` says OmniWM's handler "fired on any horizontal component". `docs/omniwm-capabilities-config.md:322` says the axis is forced **vertical** at runtime when `workspaceSwipeFingerCount` equals `fingerCount` while scroll gestures are on — which is exactly this config, four and four with `scrollEnabled = true`. So `workspaceSwipeAxis = "horizontal"` was never in effect, and the collision was on the swipe **up**, against the overview. Same fix, different mechanism; the sentence should say which.

**One stale twin.** `docs/omniwm-port.md:64-73` still describes the vertical-only demotion and calls the horizontal overlap harmless, which this PR contradicts.

**Ordering.** Land #37 first. This hands sole ownership to the engine whose arming delay that one fixes, so merged the other way round there is a window where neither side is driving.

datarip:
Thanks, you are right about the mechanism. db1157d:

- **The reason:** the comment at `bin/omacosy-wm-switch:95` now says what `docs/omniwm-capabilities-config.md:322` says. With four and four and `scrollEnabled = true`, OmniWM forces its swipe vertical, so the collision was the swipe up switching workspace under the overview. That also fits the measurement in the description better: every one of the 17 swipe-ups switched, which a diagonal-only cause would not explain. The description is corrected in the same way.
- **`docs/omniwm-port.md:64-73`:** that entry describes the current setup now (omacosy-gesture owns all four directions, OmniWM's swipe is off, and the collision was on the swipe up), and it says the 2026-08-26 vertical-only setup is gone.
- **Order:** agreed, #37 first. The description says so too.

PR #36, opened 2026-09-13
fix(bar): a window-manager switch is noticed while the bar runs

datarip:
## The symptom

Switch from AeroSpace to OmniWM while the bar is running and the workspace pills freeze on the old manager's state: the ring stays on whichever workspace AeroSpace last reported, empty workspaces stop showing their fill pill, and clicking a pill does nothing useful. Only restarting the bar fixes it.

## The WM observer cannot see the switch

`bar.swift` watches for OmniWM appearing and vanishing:

```swift
for event in [NSWorkspace.didLaunchApplicationNotification,
              NSWorkspace.didTerminateApplicationNotification] { ... }
```

`OmniWM.app` sets `LSUIElement = true`. Those notifications do not arrive for it, so `startOmniWatch()` only ever runs from the top-level call at launch — and a bar that started under AeroSpace never establishes the `workspace-bar` watch at all. It goes on shelling a now-dead `aerospace` for its workspace list.

## Measured

From a 4,000-line bar log covering many switches on OmniWM 0.6.10, macOS 27.0:

| | |
| --- | --- |
| `wm: OmniWM launched` / `quit` | **0 occurrences, ever** |
| `omniwm: watching workspace-bar` | only ever within a second of a bar start |

The decisive case: OmniWM launched fresh (a new pid) at 14:39:39 while the bar had been running for 21 minutes. The bar was demonstrably healthy and logging `frontapp` lines six seconds later — so its notification centre was alive and its `didActivateApplicationNotification` observer was firing for OmniWM specifically. The launch notification simply never came.

That asymmetry is the evidence: **activate fires for this app, launch does not.** `LSUIElement` is the explanation that fits it, inferred from the behaviour rather than from documentation.

## The fix

Reconcile the watch in the front-app observer, which does fire for OmniWM:

```swift
if omniwmActive() { startOmniWatch() } else { stopOmniWatch() }
```

Both calls are already idempotent — `startOmniWatch()` returns early unless the watch is nil and OmniWM is running, `stopOmniWatch()` returns early unless a watch exists. Cost is one `runningApplications(withBundleIdentifier:)` lookup per app activation.

The existing observer is left in place. It is correct for any future non-accessory manager, and it carries the monitor-id re-resolution and staggered rebuild retries that a switch still needs.

## Verification

Killed the watch child, activated an app, and the watch re-established with no bar restart — which was impossible before:

```
14:53:39  omniwm: watching workspace-bar     ← bar startup
14:53:44  omniwm: watching workspace-bar     ← reconcile, after a kill + app activation
14:53:44  frontapp Finder 3.56 ms
```

The reconcile line lands before the `frontapp` line, which is where it sits in the closure. Confirmed again at `14:55:33`.

Afterwards `omniwmctl watch workspace-bar` reported the correct state — workspace 3 focused holding Zed — and the pills tracked workspace changes again.

Tested on my fork against the same OmniWM version. Built from my branch rather than a clean `main` install, since my `bar.swift` carries other changes; the reconcile itself is independent of them.


datarip:
I added a second commit, 80618c9. The first commit starts the OmniWM watch after a switch, but the bar kept the display labels of the manager it started under. The two managers name the same display differently (`1` under AeroSpace, `display:1` under OmniWM), and the bar drops events for a label no bar carries. So after a switch without a bar restart, the workspace ring stayed on the old workspace.

The check is now one function, `reconcileWM()`. When the running manager changes, it also renews the labels, at 0, 1, 3 and 8 seconds, for a manager that does not answer yet. It runs on each app activation, as before, and every 2 seconds, for a switch that activates no app, such as a move between empty workspaces.

Tested on a MacBook Air M1, macOS 27.2, OmniWM 0.7.1, with `omacosy-wm-switch` in both directions and no bar restart:

| Switch | Bar log |
| --- | --- |
| OmniWM to AeroSpace | `monitor: Built-in Retina Display is now aerospace monitor 1 (was display:1)`, about 1 s into the switch |
| AeroSpace to OmniWM | `monitor: Built-in Retina Display is now omniwm monitor display:1 (was 1)`, about 1 s into the switch |

After the switch to OmniWM, the ring followed workspace 3, then 1. Before this commit, it stayed on 1 while OmniWM was on 3.

This merges cleanly with #60, which also changes `helper/bar.swift`.

paulsp94:
Thanks for the update — the blocker from the last round is fixed. `reconcileWM()` (`helper/bar.swift:3261-3276`) now tracks `labelsFromOmniWM` and runs `rebuildSurfaces()` plus `kickRebuild()` on a change, so per-monitor pills leave the old manager's id namespace. The gap itself is still real: OmniWM is `LSUIElement`, so the launch and terminate notifications can't be relied on.

**The new blocker is the timer.** `Timer.scheduledTimer(withTimeInterval: 2, repeats: true)` at `:3280` is a permanent poll, and the file header at `:25` — "nothing here polls except the clock and the weather" — is now false. It also brings a second ladder (0/1/3/8 s) beside the observer's existing one (1/3/8/15 s): if the terminate notification ever does fire, both run, and yours gives up seven seconds sooner.

**Two leftovers from last time.** `reconcileWM()` is still the first line of the activation closure (`:3286`), ahead of the front-app dedup and own-pid guard at `:3290`, so it runs on every activation including the bar's own. And the observer comment at `:3230-3235` still says "OmniWM.app appearing or vanishing is the signal".

**This needs settling together with #60.** The two merge without a textual conflict, but their logic interacts: #60 makes `rebuildSurfaces()` keep a display's last label when the manager doesn't list it yet, so an incoming manager that is still silent after your 8 s ladder leaves the *old* manager's label in place for good, and #60's own retry does not fire for that case. Between them the bar would also gain two separate 2 s loops.

The shape I'd like is one reconcile routine that owns both concerns — driven by the events that already exist (activation, the NSWorkspace observers, screen-parameter changes), plus a single guarded retry with backoff and a cap — that re-resolves monitor ids, starts or stops the OmniWM watch, and hides pills while a display is unresolved rather than showing stale ones. Either one PR, or this one rebased onto #60 with a single loop between them. Details on #60's side are in the comment there.

datarip:
Closing in favour of #60, as you suggested: one routine for both concerns. This PR's two commits are included there, rebased, and its 2 s timer is replaced by KVO on the running applications. Details and tests are on #60.

PR #42, opened 2026-09-16
fix(install): keep a real ~/.zshrc working, and out of the clone

datarip:
### The issue

`install.sh:145` makes `~/.zshrc` a symlink into the clone:

```sh
link "$REPO_DIR/zsh/zshrc"           "$HOME/.zshrc"
```

Your `~/.zshrc` stops being a file of yours and becomes a pointer at a tracked file inside the repository. Two things follow from that, and both hit ordinary users.

### What it costs you, 1: your own shell config stops loading, and nothing says so

If you had a real `~/.zshrc` — nvm, pyenv, your aliases, your PATH — `link()` moves it to `~/.zshrc.bak.<stamp>` (`install.sh:95-99`) and puts the symlink in its place. Nothing ever sources that backup again; `install.sh`, `uninstall.sh` and `zsh/zshrc` contain no reference to it. No message tells you either. Your next terminal is simply missing everything you had, and the only clue is a dotfile with a timestamp in your home folder.

### What it costs you, 2: your clone goes dirty, and updates stop

Installers append to `~/.zshrc` as a matter of course — nvm, conda, rustup, Homebrew. Through the symlink, that append lands in the repository:

```
$ ls -l $HOME/.zshrc
$HOME/.zshrc -> <clone>/zsh/zshrc
$ printf '\nexport NVM_DIR="$HOME/.nvm"\n' >> $HOME/.zshrc
$ git -C <clone> status --short
 M zsh/zshrc
```

`bin/omacosy-update:38-41` then refuses to run:

```
the clone has local changes — commit, stash or discard them first
```

It names `zsh/zshrc` as modified — a file you never opened. Updates stay blocked until you work out that your `~/.zshrc` and that file are the same thing.

### What this changes

`~/.zshrc` becomes a real file: a four-line stub that sources the clone's `zsh/zshrc`.

- a real `~/.zshrc` you already had is copied to `~/.zshrc.local`, which `zsh/zshrc` sources, so your tooling keeps loading. It is still backed up as well.
- an existing `~/.zshrc.local` is **never** merged into. Which lines win is not the installer's call. Install says so and points at the backup.
- a symlinked `~/.zshrc` (dotfiles managers, stow) is recorded as `prior-symlink`, so `restore()` puts it back at uninstall, unchanged.
- the stub is written once. Lines appended after that are never rewritten, and they land in your `~/.zshrc`, not in the clone.
- if the clone moves, one line of the stub is edited in place rather than the file being rewritten, so appended lines survive.
- `uninstall.sh` removes the stub only if it is byte-for-byte what was written. If you added lines it keeps the file as `~/.zshrc.omacosy-stub.bak.<stamp>` and says where. It removes `~/.zshrc.local` only if the install created it and it still equals the restored `~/.zshrc`.

### The three commits

1. **the stub** — the change above.
2. **appended lines survive a clone move** — repointing used to copy the whole `~/.zshrc` over the saved original, so uninstall took appended lines for its own and deleted them. The saved copy now gets the same one-line edit.
3. **omacosy's own copy is not mistaken for your config** — this one guards commits 1 and 2 on an upgrade path; it is not a defect that exists today. A clone under `~/Documents`, `~/Desktop` or `~/Downloads` is unreadable by launchd services, so `install.sh` copies configs instead of symlinking. A **stock** install from such a clone leaves `~/.zshrc` as a copy of `zsh/zshrc`, marked `copied-config`. Installing this branch over it, `zshrc_setup` would take that copy for your config and put it in `~/.zshrc.local` — which `zsh/zshrc` sources, so the file would source itself. Measured on macOS 27, zsh 5.9: the shell produced no output and was killed after 20 seconds; the same run without that file finishes at once. `zshrc_setup` now makes the same manifest check `link()` already makes at `install.sh:95`. In `uninstall.sh`, `zshrc_unstub` moves above the copied-config sweep, because that sweep does `rm -rf` on `~/.zshrc` and would otherwise take the stub and any appended line with it. A fresh install cannot reach this: this branch never calls `link()` for `~/.zshrc`.

### Test

A local sandbox suite slices the `~/.zshrc` parts out of `install.sh` and `uninstall.sh` — the real lines, not a copy — and runs them under bash 3.2 in a throwaway `HOME`, for nine starting states:

| | |
| --- | --- |
| A | no `~/.zshrc`, no `~/.zshrc.local` |
| B | a real `~/.zshrc` with tooling |
| C | `~/.zshrc` symlinked into a dotfiles repo |
| D | stock omacosy already installed over a real `~/.zshrc`, then `~/.zshrc.local` made |
| E | stock omacosy already installed over a dotfiles symlink |
| F | this build already installed, plus `~/.zshrc.local` |
| G | a relative stow symlink, plus `~/.zshrc.local` |
| H | a real `~/.zshrc` **and** a `~/.zshrc.local` |
| I | stock omacosy installed from a clone under `~/Documents` — copy mode |

Each state checks that a new zsh loads what it should, and that after uninstall the `HOME` is **byte-identical** to before the first install — same files, same modes, same contents, same symlink targets. Three further checks cover cases no starting state reaches: an installer appending to the stub, the same after a clone move, and the copied-config sweep against the stub.

Result on this branch: **9 of 9 states, 82 checks, 0 failures**. The stock installs are built from `upstream/main` at the time of the run.

Built and run daily on one machine: MacBook Air M1, macOS 27, a single 1440x900 display with no notch, tested under AeroSpace and OmniWM.


paulsp94:
Mechanism is sound — the stub is idempotent across re-installs, the in-place repoint covers `zshrc-stub.orig` so appended lines are not misattributed, and uninstall's unstub runs before the sweep behind a `cmp -s` gate. One blocker and one note.

**Blocker, one line.** `install.sh:157`'s case arm is `*omacosy*)` only, while `link()` at `:88` matches `"$src" | *omacosy*)`. For a clone whose path does not contain "omacosy", upgrading from a stock install records a *second* `prior-symlink ~/.zshrc <clone>/zsh/zshrc`, and `uninstall.sh:108` takes `tail -1` — so uninstall relinks into the clone and the user loses their own dotfiles link. `"$repo/zsh/zshrc" | *omacosy*)` closes it.

**Worth a comment rather than a change.** `install.sh:176` truncates with `>` when `copied-config $HOME/.zshrc` is marked, with no backup, so lines appended to a copy-mode `~/.zshrc` are lost on upgrade. That matches main's `link()` behaviour, so not a regression — just undocumented.

Also needs a rebase: #39 and #40 both landed in `install.sh` after you opened this.

The problem itself is confirmed on main: `install.sh:145` links over a pre-existing `~/.zshrc`, the backup at `:95-99` is never re-sourced, and `zsh/zshrc:20` sources only `~/.zshrc.local`.

datarip:
Thanks. Rebased onto `main` (no conflict with #39/#40), and 1e8960e answers the review:

- **Blocker:** the case arm in `zshrc_setup` is `"$repo/zsh/zshrc" | *omacosy*)` now, the same test as `link()`.
- **The copy-mode truncation:** commented at step (d): a `copied-config` `~/.zshrc` gets no backup before the stub overwrites it, as with `link()`.

Tested with the functions from the branch in a temporary HOME: a clone at `$HOME/src/dots-clone` (no "omacosy" in the path), `~/.zshrc` a link to `$HOME/dotfiles/zshrc`, then `link()` (the stock install), `zshrc_setup` (the upgrade), and `zshrc_unstub` + `restore` (the uninstall).

| | prior-symlink lines | `~/.zshrc` after uninstall |
| --- | --- | --- |
| before 1e8960e | dotfiles, then the clone | the clone |
| after | dotfiles | `$HOME/dotfiles/zshrc` |

PR #44, opened 2026-09-16
fix(spawn): a burst of launch chords opens one window, 13 s late

datarip:
Under OmniWM, a burst of `Super+Enter` freezes for about 13 seconds and then
opens **one** terminal window. `Super+Shift+F` does the same. Under AeroSpace
the same burst gives one window per press, immediately.

## Karabiner kills the command it is still running

OmniWM's hotkey system has no exec command, so `omacosy-karabiner-omniwm`
moves every chord that runs something into Karabiner's complex modifications.
That is the part I had wrong: **Karabiner runs one `shell_command` at a time,
and SIGKILLs the one still running when the next key arrives.**

Five fast `Super+Enter`, traced through the spawn script and sampled with `ps`
every 0.3 s:

| press | shell | outcome |
| --- | --- | --- |
| 1 | 50593 | took the lock, started its AppleScript, dead 0.2 s later |
| 2 | 50609 | dead |
| 3 | 50618 | dead |
| 4 | 50635 | dead |
| 5 | 50653 | survived — nothing followed it |

The system log confirms press 1 was not hung: its `osascript` ran from
16:13:53.208 to 16:13:53.292 and exited cleanly. The shell around it was
killed.

## SIGKILL runs no EXIT trap, so the lock outlived its owner

Press 1 held the `mkdir` lock when it died, and SIGKILL cannot be caught, so
`trap ... EXIT` never ran and `/tmp/omacosy-spawn-$(id -u).lock.d` stayed
behind with no owner. Presses 2 to 5 then waited on a process that was already
gone.

The stale-lock fallback should have covered that. It did not, for three
separate reasons:

```sh
for i in $(seq 120); do
  mkdir "$LOCK" 2>/dev/null && break
  [ "$i" -ge 100 ] && { rm -rf "$LOCK"; }
  sleep 0.05
done
trap 'rmdir "$LOCK" 2>/dev/null' EXIT
```

1. **The timer is an iteration count, and the comment above it says 5 s.** A
   turn costs `mkdir` plus `sleep 0.05` plus process overhead, about 130 ms in
   practice, so iteration 100 arrives at ~13 s, not 5 s. Measured: the steal
   fired 12.87 s after the press.
2. **Falling out of the loop runs the spawn without the lock.** After 120
   turns the loop simply ends; nothing checks whether the `mkdir` ever
   succeeded.
3. **The trap removes whatever lock it finds.** A spawn that never held the
   lock still deletes it on exit, including one another spawn is holding.

## The fix

**Detach the work.** Every chord is now emitted as
`{ <command> ; } >/dev/null 2>&1 &`. Karabiner gets a child that exits at once
and has no live pid to kill; the work continues in a process it does not know
about. This is what AeroSpace's `exec-and-forget` already does on the other
side, which is precisely why AeroSpace never had this bug.

Serialisation is not lost, and was never Karabiner's to provide. The lock in
`omacosy-spawn` is what orders a burst — and it can only do that if the spawns
live long enough to queue.

**Make the lock survive its holder.** The holder writes its pid inside the
lock directory. A waiter that finds a dead owner clears it immediately instead
of waiting out a timer. The fallback is timed on the clock (`SECONDS`) rather
than counted in sleeps, and release is owner-checked, so a spawn can no longer
delete a lock belonging to another.

The pid lands just after the `mkdir`, so a waiter that reads an empty one
re-reads after 50 ms before treating it as dead.

## Tested

| case | before | after |
| --- | --- | --- |
| holder SIGKILLed exactly as Karabiner does it | 13 s to recover | immediate |
| four spawns queued behind a killed holder | wedged | all four complete |
| 4 x `Super+Enter`, script level | 4 windows, 1.37 s | 4 windows, 1.53 s |
| 4 x `Super+Shift+F` | 3 windows, 6.33 s | 4 windows, 3.84 s |
| old command form under a kill | work lost | — |
| new command form under a kill | — | work completes |

The last two rows are the mechanism on its own: `sh -c 'cmd'` killed mid-run
loses the work, `sh -c '{ cmd ; } &'` killed mid-run completes it.

Confirmed on the keyboard afterwards: four `Super+Enter` and four
`Super+Shift+F` each give four windows, and `ps` shows every spawn living its
full ~0.4 s instead of being killed. The Finder burst's window-count baselines
step 6 -> 7 -> 8, so the lock is still sequencing them.

## Scope

Both changes are needed. The detach alone leaves a lock that any crash or kill
still strands for 13 s; the lock repair alone leaves Karabiner killing four
spawns out of five.

Every chord the generator emits is wrapped, not only the two that were
reported. The same kill applies to any command that is not instant, and
`theme-next` and `theme-bg-next` write files — being killed halfway through is
worse than running twice.

MacBook Air M1, macOS 27.0 (26A428), Karabiner-Elements 16.3.0, OmniWM 0.6.10 (build 81), IPC protocol 15.


paulsp94:
The stall is real and your arithmetic holds: `seq 120` with the steal at `i >= 100`, and a turn costs the sleep plus two forks, so about 130 ms rather than 50 — roughly 13 s to the steal, 13.2 s typical to a window, 20 s worst case. Moving the trap before the loop and checking ownership at `:77` is a genuine fix on its own; main deletes whichever lock happens to be there.

**Blocker.** `:89`'s `[ "$SECONDS" -ge "$give_up" ]` overrides the `kill -0` check at `:87`, so past 8 s a waiter `rm -rf`s a **live** holder's lock and re-mkdirs. That is reachable: five cold spawns at ~2 s each get there, and the longest legitimate hold is now the 5 s AppleScript cap from #12 plus the 2 s window wait. Put the give-up above that, around 15 s, and let the liveness check do the real work.

**Second, same loop.** `:90`'s `continue` skips both the sleep at `:93` and the counter at `:91`, so the `turns > 400` escape at `:92` is unreachable once give-up is passed, and three or more late waiters spin deleting each other's locks with no backoff.

**Two corrections for the description.** Only Super+Enter goes through `omacosy-spawn`; the browser, Finder, music and messenger chords call `open -a` directly with no lock, so they cannot show the 13 s — only the killed-holder half. And the same count-as-clock bug you are fixing still stands on the AeroSpace side at `main:109-114`, where 40 turns documented as 2 s is really about 5 s. Worth the same treatment while you are in there.

datarip:
Thanks. All four points are answered in 8b42a10 and in the description:

- **Blocker:** a live holder is no longer taken over at 8 s. The only takeover of a live holder is when its lock is over 15 s old (the 5 s AppleScript cap plus the 2 s window wait, with margin); before that, `kill -0` decides. The age comes from the lock's `pid` file, not from each waiter's own start. With per-waiter clocks, the first takeover left every other late waiter past its own give-up, and they took over the new, live lock in turn. Measured: 4 of 5 queued spawns overlapped.
- **The `continue`:** gone. Every turn sleeps 50 ms, and the "never wedge" escape is 30 s on `SECONDS`, so it can be reached.
- **The description:** it now says that on `main` only Super+Enter goes through the lock, so the browser, Finder, music and messenger chords show only the killed-holder half. Super+Shift+F joins the lock with #26.
- **The AeroSpace side:** on the clock too. I kept 5 s, what the 40 turns really took, rather than the documented 2 s. The split hint has only ever been tested against 5 s, and this PR should not change what AeroSpace does. The comment says 5 s now. Happy to go to 2 s if you prefer.

Tests (a copy of the script with the lock and state paths in a temp dir, and `open` replaced by a stand-in that logs start and end; 3 runs):

| case | first commit | 8b42a10 |
| --- | --- | --- |
| live holder for 10 s, one waiter | lock taken at 7.5 s | waits: 10.2 s |
| holder SIGKILLed | 0.40 s | 0.45 s |
| holder hung 60 s, five waiters | 4 overlapping pairs | one takeover at ~15 s, 0 overlaps |

Live, under OmniWM on macOS 27.2: 5 runs of three Super+Enter 150 ms apart all tiled as the dwindle staircase.

Note: #26 edits the same file (`--cmd` around `spawn_app`), so whichever of the two lands second has a small conflict there. I can rebase that one.

datarip:
Merge order, now checked with `git merge-tree` against current `main`: #44 conflicts only with #26 (both edit `bin/omacosy-spawn`, in either order). #28 merges cleanly with it. Whichever of #26 and #44 lands second, I will rebase that one; the combined form is #26's `--cmd` around `spawn_app` plus this lock loop.

PR #47, opened 2026-09-17
feat(bar): the icon-only pills are squares

datarip:
The bar sizes every pill the same way: ten points of padding on each side of
its contents. A pill that holds one glyph and no label therefore comes out as a
rectangle whose width follows the glyph, so four pills that read as one family
end up four different widths.

Measured with the shipped fonts, at the current `pillHeight` of 26:

| pill | today | with this change |
| --- | --- | --- |
| apple | 30.2 x 26 | 26 x 26 |
| wi-fi | 32.6 x 26 | 26 x 26 |
| bluetooth, nothing connected | 27.4 x 26 | 26 x 26 |
| activity | 30.5 x 26 | 26 x 26 |

## What it does

A `squarePills` set names the three right-cluster pills that are square when
they carry an icon and nothing else. The apple pill is the fourth and is always
one. A square pill is `pillHeight` wide, and the glyph centres on the whole
pill rather than on a ten point inset.

## What does not change

- **Neither end of the bar moves.** The apple pill keeps its left edge at
  `padLeft`. The right cluster is laid out from the right edge inwards, so the
  activity pill keeps its right edge. Only inner edges move.
- **A pill that carries a label keeps the old layout at the old width**: the
  bluetooth device count when a device is connected, and the `off` state of
  wi-fi and bluetooth.
- The weather pill is not in the set, because its glyph is part of its label.

## Tested

Built from this branch and run on a 1440x900 built-in display, macOS 27.2.

This is cosmetic, and I will not argue for it if it is not the look you want.
It is one constant and three lines, so it is as easy to drop as it is to take.

## Image

pr-square-pills


paulsp94:
Two small things and this is ready.

**Drop the set, keep the predicate.** `squarePills` (`helper/bar.swift:608`) is redundant with what it is ANDed against at `:2651`. `hasIcon && !hasLabel` already selects exactly those three: activity is icon-only by construction (`:3599`), wifi only when it is on (`:990` vs `:983`), bluetooth only with nothing connected (`:1014`), while clock, battery, volume and brightness always carry a label, and weather sets `icon = ""` so it never qualifies. The set changes no pixel today, and it is the kind of list CONTRIBUTING.md warns about — the next icon-only pill silently misses the treatment.

**One stale comment.** `:2637-2642` still says an icon-only pill is sized on the glyph's ink, which `:2652` now contradicts. A sentence fixes it.

Checked the rest and it holds up: ink centring survives, since `drawIcon` centres on `ink.midX` inside whatever box it is handed, so the off-centre bug that comment documents is not reintroduced; no clipping, the widest square glyph ink is about 12.6 inside 26; right-to-left layout stability is intact; bar height and the reserved top gap are untouched. Only the apple pill's inner edge moves, by about 4 points.

datarip:
Thanks. Both done in a953e3e:

- **The set is gone.** `hasIcon && !hasLabel` alone decides the square, so a new icon-only pill gets it with no list to update. The bluetooth count and wifi's "off" are labels, so those pills still flow icon-then-text and grow with the text.
- **The comment** now says an icon-only pill is a square, like the apple pill, with the glyph centred on its ink.

The bar builds; no pixel changes, as you found.

PR #49, opened 2026-09-18
fix(overview): the overview keeps an old theme's colour and wallpaper

datarip:
The workspace overview does not follow the current theme. It has two
faults:

- **The colour.** After a theme change, the overview keeps the colour of the
  theme that was current when it started. The bar and the window borders
  take the new theme at once.
- **The wallpaper.** The overview draws the wallpaper behind the cards
  itself. It takes the first file, in alphabetical order, from the theme's
  `backgrounds` folder. After `Super+Shift+B` that is a different picture
  from the one on the desktop. And while the overview is open, a theme or
  wallpaper change does not reach it at all.

## The colour

`omacosy-overview` is a resident daemon: the first swipe starts it, and every
later swipe only signals it to show. It reads the theme colour one time, at
start:

```swift
let accent = themeAccent()
```

`theme-set` does not restart the overview, so a theme change never reaches
it. The active card ring, the active workspace number, the hover ring, the
search box border and the drag ring all keep the old colour.

### To reproduce

With the overview closed:

1. Swipe up once, so that the daemon runs. Close the overview.
2. `theme-set gruvbox`
3. Swipe up. The rings are still the colour of the previous theme, for
   example Tokyo Night blue, while the bar and the window borders are
   gruvbox orange.

With the overview open:

1. Swipe up.
2. Press `Super+Shift+T`. The bar and the borders change. The overview
   keeps the old colour until it is closed and opened again.

On this machine the daemon had run for more than a day. The bar showed one
theme's accent and the overview showed another theme's colour, from a theme
set the day before.

## The wallpaper

To reproduce:

1. `theme-set tokyo-night`. The desktop shows the theme's first wallpaper.
2. Press `Super+Shift+B`. The desktop shows the second wallpaper.
3. Swipe up. The overview shows the first wallpaper.

`theme-set` and `theme-bg-next` both write the wallpaper they apply to
`~/.local/state/omacosy/background`. The overview did not read that link.

## The fix

For the colour:

- `showOverlay()` reads the theme each time the overview opens. `accent`
  becomes a `var`. The cost is one read of `borders.sh`, about 120 bytes,
  per swipe.
- The daemon watches `~/.config/omarchy/current`, the directory where
  `theme-set` swaps the theme link. When it changes, the overview reads the
  colour again and, if it is open, redraws the cards with `rebuildCards()`.
  `omacosy-bar` watches the same directory for the same reason, and the
  `watch()` function is a copy of the bar's, without the `create` flag it
  does not need.

For the wallpaper:

- The overview takes the picture from `~/.local/state/omacosy/background`.
  When the link is missing, it falls back to the old choice.
- The daemon watches `~/.local/state/omacosy`. When the link points to a new
  picture while the overview is open, it loads that picture behind the
  cards. Other files in that directory also wake the watcher, the
  overview's own pidfile included, so it compares the link's target first
  and does nothing when it is unchanged.

## Environment

```
macOS      27.2 (26B5086k)
Mac        MacBook Air M1, one built-in display, 1440x900, no notch
OmniWM     running; the overview code path is the same under AeroSpace
omacosy    this branch, cut from f62787a
```

With the binary built from this branch, each reproduction above now shows
the current theme's colour and the desktop's wallpaper, with the overview
closed and with it open.

**Not tested from a plain `main` install.** Installing from a `main` checkout
on this machine would replace the configs and the symlinked commands it runs
every day. The branch was cut in a `git worktree`, and only the
`omacosy-overview` binary built from it was installed and tested.


paulsp94:
Verified and good to merge. Both faults reproduce by inspection: `helper/overview.swift:531` on main is a top-level `let accent = themeAccent()`, read once for the daemon's life, and the wallpaper at `main:1246-1256` always takes the theme's alphabetically first background while ignoring `~/.local/state/omacosy/background`, which `bin/theme-set:52` and `bin/theme-bg-next:18` write.

The approach is the house one rather than a new invention: the watcher is `helper/bar.swift:3065-3086` with the same kqueue mask and re-arm, watching the directory `helper/borders.swift:592` already watches and `bin/theme-set:70` documents as the notification. Symlink swaps are handled — `ln -nsf` unlinks then recreates, and a handler landing in that gap gets nil, leaves the layer alone, and converges on the following event. Cost while hidden is two file descriptors, and `rebuildCards()` returns early when the overlay is not visible.

**One line worth adding:** `showOverlay()` never updates `shownWallpaper`, so after a missed event the guard can swallow a genuine reload while the overview is open. Assigning the resolved URL there closes it.

Minor, no action needed: the per-show `accent = themeAccent()` is redundant once the watcher runs, though keeping it as cover for the re-arm gap is defensible.

PR #50, opened 2026-09-19
fix(wallpaper): a screen connected later takes the current theme's picture

datarip:
### The issue

macOS keeps the desktop picture **per display**. `theme-set` sets the new picture only on the screens that are attached at that moment. Its own comment says so, at `bin/theme-set:43`:

```sh
# Wallpaper: first image in the theme's backgrounds/ dir, on every screen.
```

`omacosy-helper wallpaper <path>` loops over `NSScreen.screens` (`helper/main.swift:146`). A display that is unplugged during a theme change is not in that list. When it comes back, macOS restores the picture it stored for it. The desktop then shows two themes: the new one on the laptop and the previous one on the external display. No error appears anywhere.

The same happens when you change the theme with the laptop alone, shut down, connect the display and start again.

`omacosy-bar` already receives `didChangeScreenParametersNotification` (`helper/bar.swift:3288`). It rebuilds its surfaces there, but nothing applies the wallpaper again.

### The change

**`helper/main.swift`: `omacosy-helper wallpaper resync`**

- It reads the recorded picture from `~/.local/state/omacosy/background`. `theme-set` and `theme-bg-next` already write that link, so the change adds no new state.
- A screen that already shows that picture is skipped.
- A screen that shows another picture **from the themes directory** gets the recorded picture.
- Any other picture stays. That is a picture the user set in System Settings, and resync never overwrites it.
- It prints one line: `wallpaper resync: 1 of 2 screen(s) reset to <path>`, plus `, 1 kept (not an omacosy picture)` when it kept one.
- It finds the themes directory by the rule `theme-set` uses: `theme-set` is in `<repo>/bin`, and the themes are in `<repo>/themes`.

**`helper/bar.swift`: when it runs**

- When the number of screens goes up. The first run starts at once, so the old picture shows only briefly.
- Once at startup, for a display that was connected while the machine was off.
- Each time twice, 4 s apart. A picture set before macOS finishes adopting a display does not stick. The second run costs nothing when the first worked, because resync skips a screen that already agrees.
- It writes the result line to `/tmp/omacosy-bar.log`.

### Cost

- No new daemon, timer or watcher. Nothing runs until a display is added, and at startup.
- A run is one short process: 0.04 to 0.06 s, and about 15 MB peak memory while it runs.

### Test

The helper was built from this branch and run against two displays, a MacBook Air M1 and an LG UltraFine. A script sets a picture on the second display, then checks what resync does. 19 of 19 checks pass:

- both screens agree: `0 of 2` reset, no screen moves
- screen 2 on another stock theme picture: `1 of 2` reset, screen 2 back on the recorded picture
- screen 2 on a macOS system picture: `0 of 2` reset, `1 kept`, the picture stays
- no recorded picture: exit 1, and the message says why
- the usage line names `wallpaper <path> | wallpaper get | wallpaper resync`

A script cannot send `didChangeScreenParametersNotification`, so the bar part was tested by hand, on the same two displays, with my daily build. That build carries the same bar hook and a larger set of changes:

- **reconnect:** unplug, change the theme, plug back in. The display shows the old picture briefly, then the new one.
- **cold start:** change the theme with the laptop alone, shut down, connect the display, start. The log shows the first run at login did not stick and the second run 4 s later did: `1 of 2 screen(s) reset` both times, and both screens then agree.
- **a picture set by hand:** set a system picture on the display in System Settings, unplug and plug back in. The log shows `1 kept (not an omacosy picture)`, and the picture stays.

Built and run daily on a MacBook Air M1, macOS 27, with the built-in display and an LG UltraFine.


paulsp94:
Verified and good to merge. The gap is real: `bin/theme-set:43-57` sets only the screens attached at the time, and nothing re-applies afterwards — the only wallpaper callers in the repo are theme-set, theme-bg-next and install.sh. The existing screen-parameters handler rebuilds surfaces, re-applies the shade and runs ws-collapse, but never touches the picture.

Good choices worth naming: it hooks that existing handler rather than adding a second observer; the separate counter is justified, since `monitorCount` only moves after the grace; the forks are on a utility queue, not a render path, and the same handler already forks ws-collapse; and it is gated on the display count *increasing*, so resolution changes, arrangement changes and undocking all do nothing.

**One sturdiness nit.** The theme roots come from `~/.local/bin/theme-set` (`helper/main.swift:71-81`), so if that symlink is missing the resync silently no-ops. Deriving them from the `current/theme` symlink the rest of the stack already follows would remove that dependency.

**One accuracy nit for the description.** The clobber protection is narrower than stated: `helper/main.swift:191-194` skips screens that already agree and leaves anything outside `themes/` alone, but a picture the user hand-picked *from* `themes/` in System Settings is still overwritten.

PR #51, opened 2026-09-19
fix(borders): the ring goes with the window it marks

datarip:
Finder answers Super+Shift+F and opens a window of its own. A terminal file
manager is the other half of that: it tiles like every other window, wears the
theme's colours, and keeps the keyboard.

**Why yazi.** A single binary, previews through small optional helpers, and
drag-and-drop support already merged upstream on kitty's OSC 72 protocol
([yazi#4005](https://github.com/sxyazi/yazi/pull/4005),
[docs](https://yazi-rs.github.io/docs/dnd/)). Ghostty has an open request for
that protocol ([discussion #12851](https://github.com/ghostty-org/ghostty/discussions/12851));
the day it ships, the same window gains drag and drop into Mail, a browser or
Slack with no change here.

**One script, because the chord exists twice.** AeroSpace's config and, under
OmniWM, Karabiner. `bin/omacosy-files` reads `TERMINAL` from `apps.conf`, the
resolved app choice the other chords already use, and falls back to Ghostty.
Only Ghostty is told to run a command (`ghostty -e yazi`), detached because that
process lives as long as the window. Any other terminal opens a window and you
type `yazi`. A missing yazi shows a notification rather than failing silently,
since a keybinding has no terminal to print into.

**Dependencies, kept small.** `poppler` for PDF, `resvg` for SVG, `sevenzip` for
archives, `fd` for search, and the symbols-only Nerd Font for icons: about 30 MB.
Video thumbnails and raw photos need `ffmpeg-full` and `imagemagick-full`, which
are keg-only and pull in around 160 dependencies, so the Brewfile names them in a
comment instead of installing them for everyone.

**Super+Shift+Y was free**: no `y` binding in `aerospace.template.toml`, and none
among the bound keys in the Karabiner rules.

**Tested** on macOS 27 with Ghostty: the chord opens a yazi window under both
window managers, repeated presses open one window each, the missing-yazi path
shows the notification, and with no `apps.conf` the launcher falls back to
Ghostty.


paulsp94:
There is a real bug here, but it is not quite the one the title describes, and the mechanism needs rework.

**What is actually broken.** Moves and resizes already ride the existing notifications and glide tick by tick (`helper/borders.swift:512-520` on main), so the ring does not lag them. What does happen is the ring **outliving** a window that is closing, quitting or minimising: nothing hides it until DESTROY lands and the 0.35 s miss debounce at `main:468-471` expires, so it sits at full colour over a fading window. That part is genuine, and worth fixing. No overlap with #11, which fixed *which* window is ringed rather than *when* the ring goes.

**The 20 Hz timer is the blocker.** `Timer(timeInterval: 0.05)` runs forever, ten times the 0.5 s heartbeat this very file calls "the only poll left" — and the diff quietly rewords that comment rather than answering it. CONTRIBUTING.md:7-10 allows a timer only as a guarded safety net, and DESTROY and VISIBILITY notifications already exist to react to.

**Two heuristics misfire.**
- Treating `kCGWindowIsOnscreen != true` as gone: AeroSpace drags windows through offscreen stash frames during workspace switches (`main:169-171`). The watcher then hides the ring, sets `justHid`, and raises the stillness gate from 0.08 s to 0.12 s — reintroducing the post-switch lag that tuning removed.
- Flagging any window under 80% of its all-time peak within a second of *any* minimise: minimising one tile makes AeroSpace retile the rest, so shrinking neighbours get flagged and the ring drops. The alpha rule catches apps that animate their own opacity.

**Suggested shape:** drop the timer, subscribe to DESTROY and VISIBILITY plus one short (~100 ms) one-shot recheck, and match the destroy or minimise payload's window id instead of guessing from area and alpha.

**One process note:** the description here is for a different change entirely — the yazi launcher from #52, down to `bin/omacosy-files` and the Brewfile — while the diff touches only `helper/borders.swift`. So the "tested on macOS 27" claim does not cover this code. Also `EVENT_WINDOW_MINIMIZE = 1327` is marked "verified by probe" with no probe anywhere in the repo.

datarip:
Thanks, and sorry about the description: it was #52's. It is replaced now, and the title says what the bug is. a216a5f reworks the mechanism along your lines:

- **The real bug:** the description now names the ring outliving a window that closes, hides, quits or minimizes. Moves and resizes are not part of it.
- **The 20 Hz timer:** the permanent one is gone. Hide and quit now come from events matched on the payload's window id (816, 804, 1326), as you suggested. A minimize announces itself with 1327, which starts a bounded check of the ringed window only (20 ms, at most 0.7 s). Your single ~100 ms recheck came too early: the genie starts 150-400 ms after 1327, and a check at +60 and +150 ms left the ring on for ~240 ms.
- **Close is the exception:** `docs/probes/ring-events.swift` shows nothing at all during the ~250 ms fade, with every code from 1 to 2400 registered. With events alone, the ring stayed 165 ms (median) over a closing window. So the opacity watch stays, but guarded: it runs only while the keyboard or mouse was used in the last five minutes, and the heartbeat arms and disarms it. Nobody at the Mac, the daemon's CPU drops from 1.70 % to 1.31 %.
- **The two guesses:** the "not on screen" test and the 80% area test are both gone.
- **The probe:** `docs/probes/ring-events.swift` is in the branch; `EVENT_WINDOW_MINIMIZE` and the two new codes point at it.

The benchmark table is in the description.

paulsp94:
Thanks for the rework. The event-matched hide and quit (816/804/1326 on the payload's window id) and the bounded minimize check are exactly what I was after.

One thing brings back what we wanted gone: `armRingWatch` (`helper/borders.swift`, called from the 0.5 s safety timer) starts a 20 Hz `ringWatch` timer whenever there has been keyboard or mouse input in the last 300 s. At the desk that is all the time, so in practice it is the permanent 20 Hz poll again, only paused when you walk away.

Since hide, quit and minimize now come from events, what is the 20 Hz watch still catching? If it is the close case (a window closed with Cmd-W or its red button, before DESTROY lands), could that use the same shape as minimize? Start a short bounded check (20 ms, at most 0.7 s) when the event arrives, instead of polling in case one comes. If there is a close path with no event at all, please name it in the comment and we can decide whether it is worth a timer.

The 2 ms sampler in `docs/probes/ring-events.swift` is fine where it is. It is a probe, not the daemon.

datarip:
Thanks. You were right that the watch was the permanent poll again. It is gone in 9c37a55, and the close case now has an event after all, just not from the window server.

**What the 20 Hz watch was still catching:** only the close fade. Checked again with the probe in `all` mode (every code 1-2400, 3 runs): during a close, the window server sends nothing until 816, 804 and 1326 at the end of the ~250 ms fade.

**The event that does exist:** the app's own Accessibility notification. `docs/probes/ax-close.swift` (new) times it against the fade: `kAXUIElementDestroyed` arrives at the start of the fade, up to 25 ms before it, and 235-275 ms before 816, for Cmd-W, the red button and an AppleScript close, 3 runs each.

OmniWM itself observes the same `kAXUIElementDestroyedNotification` on every managed window (`AppAXContext+Callbacks.swift`). Measured on three Cmd-W closes, OmniWM's own border leaves 107-140 ms after the fade starts; the ring with this change leaves 12-47 ms before it.

**So the ring now:**
- asks the ringed window's app for `kAXUIElementDestroyed` on that one window, and hides the moment it arrives. The AX calls run on a background queue with a 0.25 s messaging timeout, per CONTRIBUTING. Nothing runs between events;
- without the grant, falls back to your shape: a click starts a 20 ms check for at most 0.7 s. A Cmd-W close then leaves on 816, about 250 ms late. That is the one path with no event the daemon can see without a permission, and it is named in the comment.

**Measured** with the same benchmark as before, 10 runs each, in my fork's build of the same change: close −36 / −27 ms and quit −51 / −40 ms (median / worst, negative = the ring leaves before the window does), hide 11 / 69 ms, minimize 16 / 93 ms. Idle CPU fell from 1.31 % to 0.93 % of one core. The description has the full table.

**The cost is a new permission for `omacosy-borders`,** and I would rather say it plainly. It is the same class of grant as `omacosy-bar`, `omacosy-gesture` and `omacosy-ffm`, and it is used read-only: the app's window list and one notification on one window. It never performs an action, never writes an attribute, and never reads a window's contents. The code comment says so, and so does a new line in the README under "What it does not do". The grant is optional: refused, the ring still works as described above. If you would rather keep the daemon permission-free, I can drop the Accessibility half and keep only the click fallback.

datarip:
Follow-up with the measurements I had left out, and one correction of context.

**All ring measurements in one table.** MacBook Air M1, macOS 27.2, TextEdit on an empty workspace. Time from the moment the window starts to leave until the ring's `hide`, median / worst in ms. A negative time means the ring left before the window did.

| | 20 Hz watch (previous version) | this PR, with the grant | **this PR, with the grant** | this PR, without the grant | OmniWM's own border |
| --- | --- | --- | --- | --- | --- |
| build | my fork | my fork | **this branch** (`main` + #51) | my fork, signed under another name, never granted | OmniWM 0.7.2 |
| window manager | OmniWM | OmniWM | **AeroSpace** | OmniWM | OmniWM |
| runs | 10 | 10 | **5** | 10 | 3 |
| close | 54 / 92 (242 idle) | −36 / −27 | **−35 / −26** | 192-232 / 257 | +107 to +140 (Cmd-W) |
| quit | 28-33 / 35-43 | −51 / −40 | **−44 / −32** | 35 / 45 | — |
| hide | 3-7 / 30-33 | 11 / 69 | **40 / 65** | 3-4 / 63 | — |
| minimize | 25-67 / 75-102 | 16 / 93 | **52 / 69** | 5-59 / 98 | — |
| what hides it on a close | the 20 Hz watch | `kAXUIElementDestroyed` | **`kAXUIElementDestroyed`** | 816, or the heartbeat's tick | — |

The worst hide times include the benchmark's own workspace switch. A first AeroSpace round, started seconds after switching window manager, is left out: the lone window filled the display, so the ring was not drawn around it at all.

**The context I owed you:** on `main`, this ring runs only under AeroSpace. Under OmniWM, `main` parks it and OmniWM draws its own border. The AeroSpace column is the build in this PR, where it matters on `main`. The OmniWM columns come from my fork, where the ring also runs under OmniWM, and that is where the comparison with OmniWM's own border in my last comment was made.

**Another app, for the report's timing:** Firefox has no close fade. Its window goes at once, and `kAXUIElementDestroyed` comes 1-15 ms before 816 (3 closes by Cmd-W, 3 by the red button). I could not get Safari or VS Code to open exactly one window under the script, so they are not measured.

PR #52, opened 2026-09-20
feat(files): yazi on Super+Shift+Y, under both window managers

datarip:
Finder answers Super+Shift+F and opens a window of its own. A terminal file
manager is the other half of that: it tiles like every other window, wears the
theme's colours, and keeps the keyboard.

**Why yazi.** A single binary, previews through small optional helpers, and
drag-and-drop support already merged upstream on kitty's OSC 72 protocol
([yazi#4005](https://github.com/sxyazi/yazi/pull/4005),
[docs](https://yazi-rs.github.io/docs/dnd/)). Ghostty has an open request for
that protocol ([discussion #12851](https://github.com/ghostty-org/ghostty/discussions/12851));
the day it ships, the same window gains drag and drop into Mail, a browser or
Slack with no change here.

**One script, because the chord exists twice.** AeroSpace's config and, under
OmniWM, Karabiner. `bin/omacosy-files` reads `TERMINAL` from `apps.conf`, the
resolved app choice the other chords already use, and falls back to Ghostty.
Only Ghostty is told to run a command (`ghostty -e yazi`), detached because that
process lives as long as the window. Any other terminal opens a window and you
type `yazi`. A missing yazi shows a notification rather than failing silently,
since a keybinding has no terminal to print into.

**Dependencies, kept small.** `poppler` for PDF, `resvg` for SVG, `sevenzip` for
archives, `fd` for search, and the symbols-only Nerd Font for icons: about 30 MB.
Video thumbnails and raw photos need `ffmpeg-full` and `imagemagick-full`, which
are keg-only and pull in around 160 dependencies, so the Brewfile names them in a
comment instead of installing them for everyone.

**Super+Shift+Y was free**: no `y` binding in `aerospace.template.toml`, and none
among the bound keys in the Karabiner rules.

**Tested** on macOS 27 with Ghostty: the chord opens a yazi window under both
window managers, repeated presses open one window each, the missing-yazi path
shows the notification, and with no `apps.conf` the launcher falls back to
Ghostty.


paulsp94:
The chord really is free — no `y` binding in the AeroSpace template, none in the Karabiner generator, none in the Ghostty config — and the window-manager symmetry is there. Four things stop it, though, and the first two are the substantial ones.

**Six mandatory dependencies for an optional feature.** `bin/omacosy-files:22-27` guards for yazi being absent and exits with a notification, which is the right instinct — but `Brewfile:29-33,36` then adds six hard entries that `install.sh:62` installs for *everyone*. Those two decisions contradict each other. `font-symbols-only-nerd-font` also looks redundant beside the entry already at `Brewfile:23`.

**It bypasses the spawn path.** `bin/omacosy-files:40-44` runs the terminal directly, so under OmniWM it skips `preselect-for-focused` (`bin/omacosy-spawn:85-88`) and the window lands above or right — exactly the spiral `omacosy-spawn` exists to prevent — and under AeroSpace it skips the split-hint wait at `bin/omacosy-spawn:103-114`.

**The fallback contradicts the docs.** `bin/omacosy-files:46` falls back to `omacosy-spawn "$TERMINAL"`, a bare terminal with no yazi in it, while `README.md:323` says the chord opens yazi.

**It survives uninstall.** `uninstall.sh:152` names each bin symlink individually and `omacosy-files` is not among them.

Bigger question first, though: nothing in the tracker asks for this, and the case in the description rests on a Ghostty feature "the day it ships". Against the house rule about speculative features, I would rather close it than carry it. If you want it anyway, the shape that would fit is no Brewfile entries, the launch routed through `omacosy-spawn`, and the chord only bound when yazi is actually present.

datarip:
Thanks. I would like to keep it, so it is reshaped along your lines and made opt-in: 5a63f2e (rebased on `main`).

- **Six mandatory dependencies:** none now. The Brewfile is untouched. `./install.sh --yazi` installs yazi and its helpers only on request, only what is missing, and records them for `uninstall.sh`. The duplicate font cask is gone.
- **The spawn path:** `omacosy-files` goes through `omacosy-spawn --cmd`, so OmniWM's `preselect-for-focused` and AeroSpace's split-hint wait apply. `--cmd` is #26's, byte for byte.
- **The fallback:** gone. Without yazi or Ghostty, it posts a notification and opens nothing.
- **Uninstall:** `omacosy-files` is removed, by the same "link into a clone" rule as the loop.
- **Bound only when present:** yes, under both managers. `install.sh` enables or drops the AeroSpace line, and the Karabiner generator adds the rule only when yazi is found.

On it being speculative: I use it daily. With the flag it costs nothing to anyone who does not ask for it. The merge order with #26, #28, #44 and #59 is in the description; #59 and this both add the first flags to `install.sh`, and I will combine them in whichever lands second.

datarip:
One addition, 2nd commit: `--yazi-full` for video and raw-photo previews (`ffmpeg-full`, `imagemagick-full`, linked over any plain build). It is a separate flag because of its size (~160 dependencies). The symbols font stays out: as you said, the Nerd Font in the Brewfile and Ghostty's built-in symbols already cover the icons.

paulsp94:
Thanks. The opt-in shape is right, and the four earlier points are all fixed: no Brewfile change, launch through `omacosy-spawn --cmd`, the chord bound only when yazi is present, and uninstall removes the link.

Two things before it can go in:

- **`--yazi-full` replaces the user's own ffmpeg.** `install.sh:124-128` runs `brew link --overwrite` for `ffmpeg-full` and `imagemagick-full` over a plain `ffmpeg`/`imagemagick` that was already there. `uninstall.sh` removes the -full packages but never relinks the plain ones, so after an uninstall the user's `ffmpeg` is gone from `PATH`. Either skip the link when the plain formula is installed and say so, or record it in the manifest and have uninstall run `brew link` on the plain one again.
- **Rebase:** it conflicts with main now that #59's `--aerospace`/`--omniwm` flag parsing has landed in `install.sh`.

Minor: the yazi block runs after the `comm`-based marking (`:105-108`), so dependencies it pulls in are never recorded, and uninstall leaves them behind. Moving the block before the marking would fix that.

datarip:
Thanks. All three are done in cc84123, on top of a rebase onto `main`:

- **ffmpeg.** When a plain `ffmpeg` or `imagemagick` is already installed, `--yazi-full` says so and records `brew-relink <name>` in the manifest. `uninstall.sh` then runs `brew unlink` and `brew link --overwrite` on the plain one, after it removes the `-full` one. The unlink is there because Homebrew can still count the plain keg as linked, and `brew link` then does nothing. I kept the link over the plain one, because without it yazi finds the plain build and has no video or raw-photo previews.
- **Rebase.** One parser in `install.sh` now takes `--aerospace`, `--omniwm`, `--yazi` and `--yazi-full`, and the usage lists all four. `bin/omacosy-spawn` took #44's wall-clock wait with the `--cmd` call, the same resolution as in #26. The two branches still merge cleanly together.
- **Marking.** The yazi block moved before the `comm` snapshot, so its dependencies are recorded. Its own `mark` calls are gone, because the snapshot now records those packages too.

Tested with a stand-in `brew` on the two new blocks only: plain `ffmpeg` present, both plain ones present, neither, and `--yazi` alone. I did not install `ffmpeg-full` on a real Mac for this.

PR #55, opened 2026-09-21
docs(readme): say where to run uninstall.sh and what to do after it

datarip:
## The problem

"Back to a normal Mac" gives `./uninstall.sh` without the folder to run it
from. A user who does not know where `install.sh` put the repo cannot run it
as written.

The steps to do after the script are only in its final output: log out to get
the native menu bar back, delete the repo, and delete the privacy entries by
hand. A user reads the README before they uninstall, but sees that output only
after the script has run.

## The solution

- The command becomes `cd ~/.local/share/omacosy && ./uninstall.sh`, the
  folder the install command in the README clones to.
- A short "After the script" list gives the three steps from the script's
  final output.
- A link to "Parking the setup" helps a user who wants to stop omacosy
  without removing it.

README only. No code changes.

## Verified

- The path is the same as in the clone command at the top of the README.
- The three steps are the same as the notes `uninstall.sh` prints when it
  finishes.
- The `#parking-the-setup` link opens the existing section.


README.md
```diff
@@ -554,6 +554,19 @@ exactly that. Tools and settings you had before omacosy are never
 touched. Pre-manifest installs fall back to a conservative teardown
 that leaves all Homebrew packages in place.
 
+After the script:
+
+1. Log out and back in. The native menu bar comes back fully only then.
+2. Delete the repo, which the script keeps:
+   `rm -rf ~/.local/share/omacosy`.
+3. Optional: in System Settings → Privacy & Security, delete the
+   Accessibility, Input Monitoring and Screen Recording entries that
```
paulsp94, line 563: `uninstall.sh` prints five: Location and Bluetooth as well. Both are already in the permissions table at README:79-80, so this step just needs to match — or point at that table instead of re-listing.

README.md
```diff
@@ -554,6 +554,19 @@ exactly that. Tools and settings you had before omacosy are never
 touched. Pre-manifest installs fall back to a conservative teardown
 that leaves all Homebrew packages in place.
 
+After the script:
```
paulsp94, line 557: The script prints three more things to deal with that this list drops: AeroSpace in Login Items, the Claude dictation shortcut, and Homebrew itself.

datarip:
Thanks, both fixed in 6cc2ee3:

- **Permissions** (`:563`): the step points at the permissions table now, so it names all five and cannot fall out of step with it.
- **The three missing items** (`:557`): AeroSpace in Login Items, the Claude dictation shortcut, and Homebrew are steps now.

With #67 merged, the section also names the backup folder and the new terminal window. The steps are numbered in the order to do them; the logout is last.

paulsp94:
Thanks. The permissions pointer and the three new steps read well. Two last things:

- **`cd ~/.local/share/omacosy && ./uninstall.sh`** only works for a clone in the default place. Anyone who cloned elsewhere gets `cd: no such file or directory`, and `&&` then skips the uninstall. `uninstall.sh` finds its own clone from its path, so the path to it is all it needs. I'd show `~/.local/share/omacosy/uninstall.sh` and add "(or `<your clone>/uninstall.sh`)", or keep `./uninstall.sh` with "from your clone".
- **Step 2, "Delete the repo",** reads as required, but nothing breaks if the clone stays, and keeping it makes a reinstall a single command. Please mark it optional. The hard-coded path in `rm -rf ~/.local/share/omacosy` has the same default-location problem as the first point, and an `rm -rf` is the worst place to be wrong.

datarip:
Thanks. Both points are fixed in d8b4052:

- **The command.** It is now `~/.local/share/omacosy/uninstall.sh`, with a line under it for `<your clone>/uninstall.sh`. No `cd`, so no `&&` to skip the uninstall.
- **Step 2.** It is optional now. It says that nothing breaks if the clone stays and that a reinstall is then one command. The `rm -rf` names `<your clone>` and gives the default path after it.

I also rewrote the description. It still described the first version.

PR #56, opened 2026-09-21
fix(install): a re-run keeps OmniWM when it is the window manager

datarip:
Run `./install.sh` again on a machine that uses OmniWM, and it switches
half of the machine back to AeroSpace. `omacosy-update` runs `install.sh`,
so every update does this.

## What goes wrong

`omacosy-wm-switch omniwm` changes four things outside the repo.
`install.sh` assumes AeroSpace and undoes each one:

| `omacosy-wm-switch omniwm` does this | `install.sh` then does this | Result |
| --- | --- | --- |
| quits AeroSpace | `open -a AeroSpace` | two window managers tile the same windows |
| adds the `omacosy-omniwm` rules to `karabiner.json` | copies `karabiner.json` from the repo | every chord that runs a command (terminal, launcher, theme, lock, the workspace slots) does nothing. Caps Lock → Super still works, so the keyboard looks fine |
| stops `omacosy-ffm` | loads `omacosy-ffm` | it fights OmniWM's own focus-follows-mouse |
| writes `config.omniwm.json` to `gesture.json` | writes `config.json` to `gesture.json` | the swipes send AeroSpace commands |

To recover, I had to quit AeroSpace, run `omacosy-wm-switch omniwm`
again, and log out.

## The fix

`install.sh` finds the window manager before it changes anything. OmniWM
is the manager when one of these is true:

- OmniWM answers its socket (`omacosy-omni active`). This covers an update
  run in a live session.
- OmniWM is the login item. `omacosy-wm-switch` moves the login item when
  you confirm a switch, so this covers a session where OmniWM is not
  running yet.

When OmniWM is the manager, `install.sh`:

- does not start AeroSpace
- writes the `omacosy-ffm` launch agent, but does not load it
- writes `config.omniwm.json` to `gesture.json`
- runs `omacosy-karabiner-omniwm install` at the end, after `karabiner.json`
  was copied

On any other machine, `install.sh` does the same as before. The check
starts with `[ -d /Applications/OmniWM.app ]`, so a machine without
OmniWM does not run it.

## Tested

On a Mac that runs OmniWM, after `omacosy-wm-switch omniwm`. I checked
the state before and after `./install.sh`:

| Check | Before | After |
| --- | --- | --- |
| AeroSpace running | no | no |
| `omacosy-ffm` loaded | no | no |
| `gesture.json` is `config.omniwm.json` | yes | yes |
| `omacosy-omniwm` rules in `karabiner.json` | present | present (35) |
| `omacosy-omni active` answers | yes | yes |

`install.sh` exits 0 and prints `OmniWM is this machine's window manager:
keeping it`. Before this change, the same run changed the first four
rows: this is the break I saw on a second Mac after `omacosy-update`.

The run used my fork's build, which holds this commit unchanged. I did
not run the AeroSpace path. Its code is the same as before, inside
the new `else`.

## Scope

One file, `install.sh`, +37/-9. No new permissions. The login-item check
uses the same System Events call as `omacosy-wm-switch`. It runs only when
OmniWM is installed and does not answer its socket.


paulsp94:
Verified, and this is the most valuable of your current batch — it fixes something that bites on every update, not an edge case. On main, `install.sh:481-484` starts AeroSpace unconditionally, `:165` overwrites `karabiner.json` and drops the `omacosy-omniwm` exec rules, `:309` loads the ffm agent, and `:395` copies the AeroSpace gesture config over the OmniWM one. Since `bin/omacosy-update:83` re-runs install.sh, an OmniWM machine gets a second window manager, dead exec chords and repointed swipes every single time it updates.

The probe is the house idiom exactly — the same `omacosy-omni active` line as omacosy-float, omacosy-toggle, omacosy-ws, omacosy-cycle, omacosy-focus-guard and omacosy-ws-collapse — and the login-item fallback reuses the item `bin/omacosy-wm-switch:125-128` is the sole writer of. The AeroSpace default survives for everyone else: a fresh install, a reverted dead-man gate and `wm-switch aerospace` all take the old path. A stale socket is safe too, since connect() is refused rather than accepted.

**Two ordering nits before it lands.**

1. `install.sh:501-506` never ensures OmniWM is actually running, so the login-item path can finish with no window manager at all. The AeroSpace branch's `open -a` is idempotent; mirroring it costs one line.
2. `karabiner.json` is clobbered at `:178` and only restored at `:505`. Any `set -e` abort in between — brew, clang, macos-defaults.sh — leaves exactly the breakage this PR exists to fix. Re-injecting the OmniWM rules right after the copy closes that window.

Out of scope but worth recording somewhere: the ffm plist keeps `RunAtLoad` in `~/Library/LaunchAgents`, so it returns at the next login regardless. `omacosy-wm-switch` has the same hole.

